PR #16, opened 2025-08-06
Add DeepSeek DP+EP case

alexm-redhat:
This PR adds the DSR1 DP+EP instructions.


alexm-redhat:
@simon-mo @mgoin I have added text for DP+EP case. Feel free to add comments/modify.

PR #21, opened 2025-08-11
Add TOC

jeejeelee:


jeejeelee:
> 
> 
> There's already TOC on the right side of the page, do we still need this?

This might be more friendly for users who browse markdown directly.

PR #58, opened 2025-09-15
Add b200 benchmark numbers to Qwen3-Next.md

chelsea0x3b:
I also have some numbers for 2 speculative tokens, and for different levels of tensor parallelism on the B200. B200 can go all the way down to tp=1, but its much slower due to small kv cache.

Any requests for numbers to go in here?


simon-mo:
There were some more discussion in the slack thread

PR #83, opened 2025-10-06
Update GPT-OSS.md - Clarify minimum version requirement for --tool-ca…

cardofe:
…ll-parser openai

Update the GPT-OSS usage guide to explicitly state that the  --tool-call-parser openai flag requires vLLM version 0.10.2 or later.

The openai parser was not available in version 0.10.1, which only  supported parsers like hermes, mistral, internlm, etc. Users attempting  to use --tool-call-parser openai with version 0.10.1 will encounter:

  KeyError: 'invalid tool call parser: openai (chose from 
  { deepseek_v3, glm45, granite-20b-fc, granite, hermes, 
  hunyuan_a13b, internlm, jamba, kimi_k2, llama4_pythonic, 
  llama4_json, llama3_json, minimax, mistral, phi4_mini_json, 
  pythonic, qwen3_coder, step3, xlam })'

This commit adds:
- Version requirement notice in the Function Calling section
- Minimum version badge/note in the installation instructions
- Troubleshooting entry for users encountering the KeyError

Rationale:
The current documentation shows --tool-call-parser openai in examples  without specifying version requirements, leading to confusion when users  with version 0.10.1 cannot use this parser. Making the version  requirement explicit helps users:

1. Understand why the parser is not available
2. Know they need to upgrade to 0.10.2+
3. Choose alternative parsers if stuck on 0.10.1

Related changes:
- Update installation section to recommend 0.10.2 as default
- Add version compatibility matrix for tool call parsers
- Include upgrade instructions for users on 0.10.1

Testing:
- Verified --tool-call-parser openai works on vLLM 0.10.2
- Confirmed error message on vLLM 0.10.1
- Checked all available parsers in both versions


youkaichao:
cc @heheda12345

cardofe:
# vLLM 0.10.1
vllm 0 10 1

# vLLM 0.10.2
vllm 0 10 2

PR #141, opened 2025-12-08
add phi-4 recipe

Tahirc1:
This pr adds a recipe for Phi-4 models

### DIFF
--- a/Microsoft/Phi-4.md
+++ b/Microsoft/Phi-4.md
@@ -65,10 +65,11 @@ print("Generated text:", response.choices[0].message.content)
 ### Multimodal Example (Image + Text)
 
 > [!NOTE]  
-> To run this example, you must start the server with the `microsoft/Phi-4-multimodal-instruct` model. The command is similar to the one for the text model:
+> To run this example, you must start the server with the `microsoft/Phi-4-multimodal-instruct` model:
 > ```bash
-> vllm serve microsoft/Phi-4-multimodal-instruct --host 0.0.0.0 --max-model-len 4000
+> vllm serve microsoft/Phi-4-multimodal-instruct --host 0.0.0.0 --max-model-len 4000 --trust-remote-code
 > ```
+> This model’s multimodality support is implemented via LoRA modules, and --trust-remote-code is required to enable the execution of those components.
 
 ```python
 from openai import OpenAI


Microsoft/Phi-4.md
```diff
@@ -0,0 +1,124 @@
...
+
+print("Generated text:", response.choices[0].message.content)
+```
+
+### Multimodal Example (Image + Text)
+
+> [!NOTE]  
+> To run this example, you must start the server with the `microsoft/Phi-4-multimodal-instruct` model. The command is similar to the one for the text model:
+> ```bash
+> vllm serve microsoft/Phi-4-multimodal-instruct --host 0.0.0.0 --max-model-len 4000
```
jeejeelee: Phi-4-multimodal-instruct supports different modalities by using LoRA, so I think this section should include this information, you can refer to https://github.com/vllm-project/vllm/blob/main/examples/offline_inference/vision_language.py#L1368

Tahirc1: Thanks for pointing this out. I realized I forgot to include --trust-remote-code in the serve command, which is required for serving Phi-4-multimodal-instruct.

Microsoft/Phi-4.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+### Multimodal Example (Image + Text)
+
+> [!NOTE]  
+> To run this example, you must start the server with the `microsoft/Phi-4-multimodal-instruct` model:
+> ```bash
+> vllm serve microsoft/Phi-4-multimodal-instruct --host 0.0.0.0 --max-model-len 4000 --trust-remote-code
+> ```
+> This model’s multimodality support is implemented via LoRA modules, and --trust-remote-code is required to enable the execution of those components.
```
jeejeelee, line 72: I think we should implement this part by referring to [phi4-mm vision](https://github.com/vllm-project/vllm/blob/main/examples/offline_inference/vision_language.py#L1394) and [phi-mm audio](https://github.com/vllm-project/vllm/blob/main/examples/offline_inference/audio_language.py#L186).

Tahirc1: I’m not fully sure what you’re suggesting. Are you proposing that we reference the corresponding implementations in phi-4-mm-vision and phi-4-mm-audio?

If this involves links to specific code sections, note that line-number links can become invalid as the code evolves.

jeejeelee (approved):
Sorry for missing this PR

PR #146, opened 2025-12-09
Update GLM-V.md for AMD GPU

haic0:


haic0 (commented):
Modified some env args.

PR #158, opened 2025-12-11
Update DeepSeek-V3_1.md for AMD GPU

haic0:


DeepSeek/DeepSeek-V3_1.md
```diff
@@ -135,3 +135,51 @@ curl http://localhost:8000/v1/chat/completions \
...
+
+
+```bash
+SAFETENSORS_FAST_GPU=1 \
+VLLM_USE_TRITON_FLASH_ATTN=0 \
+VLLM_ROCM_USE_AITER=1 \
+vllm serve deepseek-ai/DeepSeek-V3.1 \
+  --tensor-parallel-size 8 \
+  --enable-expert-parallel \
+  --served-model-name ds31
```
tjtanaa, line 168: we have to remove this `--served-model-name` argument, it is conflicting with the benchmark command.

Moreover, the `vllm serve --model <model path>` expect `<model path>` to be a valid path (either a huggingface hub name, or a local path) where it can get the tokenizer values. So we cannot use custom `served-model-name`.

GoldenGrapeGentleman: Hi @tjtanaa , we have remove this argument and make a new PR https://github.com/vllm-project/recipes/pull/277 for this update, please review, thanks!

PR #272, opened 2026-03-11
Add FP8 B200 best config to MiniMax-M2.5 recipe

faradawn:
## Summary
Adds FP8 serving config fo B200 GPUs to the MiniMax-M2.5 recipe

Reference: https://github.com/SemiAnalysisAI/InferenceX/pull/757

### DIFF
--- a/MiniMax/MiniMax-M2.md
+++ b/MiniMax/MiniMax-M2.md
@@ -120,7 +120,7 @@ vllm serve MiniMaxAI/MiniMax-M2.5 \
   --enable-auto-tool-choice
 ```
 
-For 4x H200/H20 or 4x A100/A800, run like this
+For 4x H200/H20 or 4x A100/A800: Standard serving configuration
 
 ```bash
 vllm serve MiniMaxAI/MiniMax-M2.5 \


MiniMax/MiniMax-M2.md
```diff
@@ -103,9 +103,24 @@ uv pip install vllm \
...
 ### NVIDIA GPU
 
-You can use 4x H200/H20 or 4x A100/A800 GPUs to launch this model.
+You can use 2x B200, 4x H200/H20, or 4x A100/A800 GPUs to launch this model.
 
-run tensor-parallel like this:
+For 2x B200: FP8 serving on B200 GPUs with FlashInfer MoE FP8 and DeepGEMM MoE disabled:
+
+```bash
+export VLLM_USE_FLASHINFER_MOE_FP8=0
```
robertgshaw2-redhat, line 111: instead of this, can we just update the default oracle in VLLM to select a different kernel for this case?

robertgshaw2-redhat: this is where the code is to update the priority order
- https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/fused_moe/oracle/fp8.py#L57

faradawn: Hi @robertgshaw2-redhat, thank you for the comment! Can you check if this PR addresses your comment: https://github.com/vllm-project/vllm/pull/37056

PR #310, opened 2026-04-03
Fix typo in Gemma4 Usage Guide

GaurangTandon:
All the other commands use a comma to separate the two modalities. I copied this (presumably wrong) command and it errored out.

```
api_server.py: error: argument --limit-mm-per-prompt: Value image=0 cannot be converted to <function loads at 0x7f9c235a3b00>.
```


faradawn (approved):
Hi @GaurangTandon thanks for the PR. Looks good to me!

PR #315, opened 2026-04-07
Update GLM-5.1 recipe

ywang96:


sir3mat:
is it avaiable also for amd?

PR #316, opened 2026-04-07
Fix Gemma4 doc

whynotkimhari:
As written in the parser code: https://github.com/vllmproject/vllm/blob/main/vllm/reasoning/gemma4_reasoning_parser.py and https://github.com/vllm-project/vllm/blob/main/vllm/tool_parsers/gemma4_tool_parser.py.


faradawn:
Looking good. opening tokens are <|token> and closing tokens are <token|>

faradawn (approved):

PR #331, opened 2026-04-13
update glm-5.1 deployment command

JaredforReal:
add `--chat-completion-content-format=string` to avoid `chat-template` related `tool_response` issue.
see (https://github.com/vllm-project/vllm/pull/39253#issuecomment-4211850796)


ywang96 (approved):
Thank you!

PR #332, opened 2026-04-14
Add Kimi-K2.5-nvfp4 GB200 disagg recipe

kyleliang-nv:
This PR updates the Kimi-K2.5 recipe doc with GB200 disagg recipe running NVFP4 checkpoint

### DIFF
--- a/moonshotai/Kimi-K2.5.md
+++ b/moonshotai/Kimi-K2.5.md
@@ -72,6 +72,27 @@ docker run --gpus all \
     --trust-remote-code
 ```
 
+##### Blackwell — NVIDIA NVFP4 weights (`nvidia/Kimi-K2.5-NVFP4`)
+
+[NVIDIA’s NVFP4 checkpoint](https://huggingface.co/nvidia/Kimi-K2.5-NVFP4) targets **Blackwell** (for example GB200). Accept any license terms on the model card before downloading. On Blackwell MoE models, enable FlashInfer FP4 MoE kernels (same pattern as other NVIDIA FP4 MoE recipes in this repo):
+
+```bash
+docker run --gpus all \
+  -p 8000:8000 \
+  --ipc=host \
+  -e VLLM_USE_FLASHINFER_MOE_FP4=1 \
+  vllm/vllm-openai:v0.17.0-aarch64-cu130 nvidia/Kimi-K2.5-NVFP4 \
+    --tensor-parallel-size 4 \
+    --mm-encoder-tp-mode data \
+    --compilation_config.pass_config.fuse_allreduce_rms true \
+    --tool-call-parser kimi_k2 \
+    --reasoning-parser kimi_k2 \
+    --enable-auto-tool-choice \
+    --trust-remote-code
+```
+
+For higher throughput on MoE stacks you can add expert parallelism; see [Expert Parallelism Deployment](https://docs.vllm.ai/en/latest/serving/expert_parallel_deployment.html). If you hit OOM, lower `--gpu-memory-utilization` or adjust TP/EP to match your GPU count.
+
 #### AMD (ROCm)
 
 Verified on 8× MI300X/MI355X GPUs:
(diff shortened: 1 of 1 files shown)


moonshotai/Kimi-K2.5.md
```diff
@@ -116,6 +137,102 @@ vllm serve moonshotai/Kimi-K2.5 -tp 8 \
...
+NIXL / UCX (see the NixlConnector doc for transport tuning):
+
+```bash
+export UCX_NET_DEVICES=all   # or pin devices for your fabric
+```
+
+**Per process** on a host (each vLLM worker needs a **unique** side-channel port on that host):
+
+```bash
+export CUDA_VISIBLE_DEVICES=0
```
esmeetu, line 183: This env seems useless for kimi.

kyleliang-nv: We are setting it for DP, where each DP rank sets its own CUDA_VISIBLE_DEVICES. 
Explained more in the `Data parallel + expert parallel ` section. Specifically
```
 Each process gets its own --data-parallel-rank, CUDA_VISIBLE_DEVICES, and unique VLLM_NIXL_SIDE_CHANNEL_PORT on that host 
 ```

**Comment by esmeetu** [comment]
@kyleliang-nv Thanks for you contribution! Would it be possible to have prefill use `--enforce-eager`, while decode uses `compilation-config: '{"cudagraph_mode":"FULL_DECODE_ONLY"}'`?

File: moonshotai/Kimi-K2.5.md (hunk 1 above)
Comment (kyleliang-nv) [reply moonshotai/Kimi-K2.5.md:183]: Of course, another way to setup DP is to spawn 1-process-per-node, and then we don't need to explicitly set the `CUDA_VISIBLE_DEVICES`.

File: moonshotai/Kimi-K2.5.md [hunk 2]
```diff
@@ -116,6 +137,102 @@ vllm serve moonshotai/Kimi-K2.5 -tp 8 \
...
+  --data-parallel-address <leader_ip> \
+  --data-parallel-rpc-port 13345 \
+  --enable-expert-parallel \
+  --kv-transfer-config '{"kv_connector":"NixlConnector","kv_role":"kv_producer","kv_load_failure_policy":"fail"}' \
+  --trust-remote-code
+```
+
+#### Environment on a decode worker
+
+Match the **MoE FP4 / NCCL** and **UCX** exports you use on prefill unless you split them intentionally. Set **`VLLM_NIXL_SIDE_CHANNEL_PORT`** / **`VLLM_NIXL_SIDE_CHANNEL_HOST`** per worker the same way as prefill.
```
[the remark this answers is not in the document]
kyleliang-nv: The suggestion is wrong. Check https://github.com/vllm-project/vllm/blob/bcc2306cefa4179c548d3e638e7a22a88d281733/vllm/envs.py#L185

moonshotai/Kimi-K2.5.md
```diff
@@ -116,6 +137,102 @@ vllm serve moonshotai/Kimi-K2.5 -tp 8 \
...
+```bash
+export UCX_NET_DEVICES=all   # or pin devices for your fabric
+```
+
+**Per process** on a host (each vLLM worker needs a **unique** side-channel port on that host):
+
+```bash
+export CUDA_VISIBLE_DEVICES=0
+export VLLM_NIXL_SIDE_CHANNEL_PORT=<unique_port>
+export VLLM_NIXL_SIDE_CHANNEL_HOST=<routable_ip_of_this_host>  # when prefill/decode cross nodes
```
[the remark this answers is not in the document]
kyleliang-nv: The suggestion is wrong. Check https://github.com/vllm-project/vllm/blob/bcc2306cefa4179c548d3e638e7a22a88d281733/vllm/envs.py#L185

kyleliang-nv:
> @kyleliang-nv Thanks for you contribution! Would it be possible to have prefill use `--enforce-eager`, while decode uses `compilation-config: '{"cudagraph_mode":"FULL_DECODE_ONLY"}'`?

Yes. Applied the changes as you suggested.

esmeetu:
@kyleliang-nv Can you sign off commits for pass the DCO check.

PR #334, opened 2026-04-15
Add B200 GPU configuration for MiniMax-M2.7

kjiang249:


MiniMax/MiniMax-M2.md
```diff
@@ -129,6 +129,21 @@ vllm serve MiniMaxAI/MiniMax-M2.7 \
...
+
+On B200 GPUs, TP4 with the following configuration is recommended.
+
+```bash
+vllm serve MiniMaxAI/MiniMax-M2.7 \
+  --trust-remote-code \
+  --tensor-parallel-size 4 \
+  --enable-auto-tool-choice \
+  --tool-call-parser minimax_m2 \
+  --reasoning-parser minimax_m2_append_think
```
simon-mo: This is quite different from the previous configuration 
```
vllm serve MiniMaxAI/MiniMax-M2.7 \
  --tensor-parallel-size 4 \
  --tool-call-parser minimax_m2 \
  --reasoning-parser minimax_m2  \
  --compilation-config '{"mode":3,"pass_config":{"fuse_minimax_qk_norm":true}}' \
  --enable-auto-tool-choice \
  --trust-remote-code
```

Have you tested that,
* we need `minimax_m2_append_think` 
* we cannot do `--compilation-config '{"mode":3,"pass_config":{"fuse_minimax_qk_norm":true}}'` on blackwell?

simon-mo (changes requested):

MiniMax/MiniMax-M2.md
wzhao18: We can remove `minimax_m2_append_think` and add `--compilation-config '{"mode":3,"pass_config":{"fuse_minimax_qk_norm":true}}'`. Though the `fuse_minimax_qk_norm` is not yet available on latest vllm release.

simon-mo: My main concern is we have different configurations for Hpper and Blackwell and I would like to educate the users on the minimal difference between the two and explain the rationale.

wzhao18: yes we will keep them consistent. Thanks for flagging this out.

simon-mo: Thanks. please ping me on Slack when it's updated, thx

MiniMax/MiniMax-M2.md
```diff
@@ -129,6 +129,22 @@ vllm serve MiniMaxAI/MiniMax-M2.7 \
...
+
+```bash
+vllm serve MiniMaxAI/MiniMax-M2.7 \
+  --tensor-parallel-size 4 \
+  --tool-call-parser minimax_m2 \
+  --reasoning-parser minimax_m2 \
+  --compilation-config '{"mode":3,"pass_config":{"fuse_minimax_qk_norm":true}}' \
+  --enable-auto-tool-choice \
+  --trust-remote-code
+```
```
simon-mo: now this is exactly the same command as above. I think we don't need this section anymore and just add B200 is supported and the notes.

wzhao18: done! Thanks!

simon-mo (approved):

PR #354, opened 2026-04-21
Use latest image instead of gemma4

khluu:
Now that v0.19.1 has transformers v5


ywang96 (approved):

ywang96:
@khluu can you update?

Issue #366, opened 2026-04-23
Audit: env var cleanup pass across all recipes


faradawn:
We have accumulated a number of \`extra_env\` / \`base_env\` entries across recipes. This issue tracks a pass to decide which to keep, which to remove (stale or should become a vLLM heuristic), and which need better documentation.

**Goal:** confirm each decision with an engineer who knows the flag, then file follow-up PRs.

---

| Env Var | Decision | Notes |
|---------|----------|-------|
| \`VLLM_USE_FLASHINFER_MOE_FP4\` | Remove | Should be a vLLM heuristic: auto-select when variant precision is fp4/nvfp4 on Hopper. High perf impact — removing the explicit flag only matters if vLLM doesn't pick it up automatically. |
| \`VLLM_USE_FLASHINFER_MOE_FP8\` | Remove | Same family as above. The =0 workaround cases need an upstream fix; the =1 cases should be auto-detected. |
| \`VLLM_USE_FLASHINFER_MOE_MXFP4_MXFP8\` | Investigate | Only appears in gpt-oss. Unclear if this is a stable upstream flag or a one-time experiment. |
| \`VLLM_FLASHINFER_MOE_BACKEND\` | Keep | Meaningful latency-vs-throughput kernel choice that users should own. Needs better documentation on when to pick \`latency\` vs \`throughput\`. |
| \`VLLM_USE_DEEP_GEMM\` | Keep | Real throughput lift for FP8 matmul. Requires DeepGEMM install and has a warmup cost, so it should stay explicit. Worth surfacing in a performance-mode guide. |
| \`VLLM_DEEP_GEMM_WARMUP\` | Keep | Intentional escape hatch to skip slow JIT warmup at startup. No throughput effect at runtime. |
| \`VLLM_FLOAT32_MATMUL_PRECISION=high\` | Keep | Enables TF32 TensorCore path on Blackwell — medium perf gain. Currently only in guide prose; should be promoted to a structured \`hardware_overrides.blackwell\` entry across all Blackwell-capable recipes. |
| \`SAFETENSORS_FAST_GPU\` | Remove | Should be a vLLM default — no reason it isn't always-on. Speeds up model loading from disk; no inference throughput effect. |
| \`VLLM_ATTENTION_BACKEND=FLASH_ATTN\` | Keep | Model-specific correctness requirement (Qwen3-Next), not a perf choice. |
[...]

PR #367, opened 2026-04-23
chore: env var cleanup — remove stale VLLM_USE_V1 and VLLM_COMMIT flags

faradawn:
Part of the env var audit tracked in #366. Removes two flags confirmed safe to drop: `VLLM_USE_V1` (V1 is now the default engine in vLLM and the flag is a no-op) and `VLLM_COMMIT` (a one-off wheel pin in the MiniMax-M2.7 guide that was never a runtime env var pattern — replaced with the standard nightly install). Also drops the co-located `VLLM_V1_USE_PREFILL_DECODE_ATTENTION` and `VLLM_USE_TRITON_FLASH_ATTN=0` lines from the same Qwen3-235B block, which are V1-era workarounds that no longer apply.


esmeetu:
LGTM! Thanks.

PR #384, opened 2026-04-24
DSv4 Flash PD disagg with NixlConnector

NickLucche:
I ran some tests this morning using NixlConnector instead of Mooncake for PD disagg.
I just wanted to clarify in the recipes it should also be supported out of the box too (afaik NixlConnector feature-support is currently a superset of Mooncake's).


NickLucche:
cc @zhewenl

zhewenl (approved):

zhewenl:
LGTM, Nixl is actually in the tab above but it's currently broken, we are also working on cleaning up the current recipes

PR #398, opened 2026-04-27
Document YaRN ultra-long context for Qwen3.5/3.6 family

abtraore:
## Summary

The Qwen3.5/3.6 family natively serves 262K context but supports extension up to ~1M tokens via YaRN RoPE scaling, as documented on each HF model card. Only `Qwen3.5-397B-A17B` mentioned this previously, and only as a one-liner pointing back at HF.

This PR adds a **Processing Ultra-Long Texts** section to the guide of each Qwen3.5/3.6 recipe, with:

- Guidance on choosing the YaRN `factor` (`2.0` ≈ 524K, `4.0` ≈ 1M) and the short-context quality tradeoff at higher factors.
- A complete `vllm serve` example using `VLLM_ALLOW_LONG_MAX_MODEL_LEN=1` and the exact `--hf-overrides` JSON from the model cards.
- A link back to the per-model HF docs for the full parameter reference.

## Recipes updated

- `Qwen/Qwen3.6-27B`
- `Qwen/Qwen3.6-35B-A3B`
- `Qwen/Qwen3.5-397B-A17B` (also drops the obsolete one-liner under Troubleshooting now that the dedicated section covers it)

## Test plan

- [x] `node scripts/build-recipes-api.mjs` → `✓ JSON API: 80 models, 8 strategies`
- [x] Visual check on the rendered guide pages

### DIFF
--- a/models/Qwen/Qwen3.5-397B-A17B.yaml
+++ b/models/Qwen/Qwen3.5-397B-A17B.yaml
@@ -221,6 +221,23 @@ guide: |
     --enable-prefix-caching
   ```
 
+  ## Processing Ultra-Long Texts
+
+  Qwen3.5-397B-A17B natively supports `262,144` tokens. For longer inputs, apply
+  YaRN RoPE scaling via `--hf-overrides` and raise `--max-model-len`. Pick
+  `factor` to match your real workload — `2.0` covers ~524K, `4.0` covers
+  ~1M — since YaRN at higher factors degrades short-context quality.
+
+  ```bash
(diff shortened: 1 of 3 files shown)


abtraore:
Thanks @gemini-code-assist — good catch. At 1M context the KV cache (~80–120 GB) plus BF16 weights would OOM on the 2× H100 implied by the example. Pushed dda7420 swapping both 27B and 35B-A3B examples to their FP8 checkpoints so the commands are actually runnable as written.

esmeetu:
LGTM!

PR #423, opened 2026-04-29
fix(deepseek): set TP=8 for V4-Flash single_node_tp on 8-GPU nodes

dsingal0:
The auto-fit logic sets TP=1 when the model fits on a single GPU (170 GB <= 180 GB per B200). Override tp: 8 for single_node_tp so that 8xB200 and 8xH200 deployments shard across all GPUs as intended for the latency-oriented tensor-parallel-only path.


dsingal0:
@simon-mo

simon-mo:
@esmeetu

esmeetu:
LGTM! Thanks. @dsingal0

PR #434, opened 2026-05-01
Update JSON API

esmeetu:


models/deepseek-ai/DeepSeek-V3.2-Exp.yaml
```diff
@@ -26,10 +26,6 @@ model:
     - "--trust-remote-code"
   base_env: {}
 
-dependencies:
-  - note: "DeepGEMM pinned build for MQA logits (FP8 MoE kernels)"
-    command: "uv pip install git+https://github.com/deepseek-ai/DeepGEMM.git@v2.1.1.post3 --no-build-isolation"
```
mgoin, line 31: Is this newer than the deepgemm we use in-tree?

esmeetu: this is an old version.

PR #440, opened 2026-05-05
fix: correct gemma-4-26B MTP name

lianakoleva:
The model is at [`google/gemma-4-26B-A4B-it-assistant`](https://huggingface.co/google/gemma-4-26B-A4B-it-assistant), rather than `google/gemma-4-26B-it-assistant`.


santiramos27:
Thanks for the fix!

PR #444, opened 2026-05-07
fix: correct HuggingFace repo prefix in Gemma 4 MTP speculative decoding docs

lucianommartins:
## What

Replaces incorrect `gg-hf-am/` HuggingFace org prefix with the correct `google/` across all MTP assistant model references in `Google/Gemma4.md`.

## Why

The MTP speculative decoding section was pointing to non-existent HuggingFace repos under `gg-hf-am/` instead of the correct `google/` organization. This affected:

- The assistant models table (4 entries)
- The online serving `--speculative-config` example
- The offline inference `speculative_config` example

## Files changed

- `Google/Gemma4.md` — 6 occurrences fixed

### DIFF
--- a/Google/Gemma4.md
+++ b/Google/Gemma4.md
@@ -949,10 +949,10 @@ Gemma 4 supports Multi-Token Prediction (MTP) speculative decoding using lightwe
 
 | Target Model | Assistant Model | Centroids Masking |
 |---|---|---|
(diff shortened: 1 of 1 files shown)


lucianommartins:
hey @ywang96 - just fixing HF repo names here. let's merge? :)

PR #446, opened 2026-05-08
Fix Gemma 4 limit-mm-per-prompt examples

Aratako:
## Summary

Update the Gemma 4 recipe examples to use vLLM's supported dotted dictionary CLI syntax for `--limit-mm-per-prompt`.

The previous comma-separated form is rejected by vLLM argparse:

```bash
--limit-mm-per-prompt image=4,audio=1
```

This PR replaces it with dotted keys:

```bash
--limit-mm-per-prompt.image 4 \
--limit-mm-per-prompt.audio 1
```

The change is applied to both the canonical YAML recipes and the legacy Gemma4 markdown page.

Related to vllm-project/vllm#39687.

## Validation

- Verified against vLLM argparse handling that:
  - `--limit-mm-per-prompt.image 4 --limit-mm-per-prompt.audio 1` parses to `{"image": 4, "audio": 1}`
  - `--limit-mm-per-prompt image=4,audio=1` is rejected

### DIFF
--- a/Google/Gemma4.md
+++ b/Google/Gemma4.md
@@ -173,8 +173,8 @@ docker run -itd --name gemma4-rocm \
 
 - Set `--max-model-len` to match your actual workload. The default context length can be very large; reducing it saves memory for KV cache.
 - Use `--gpu-memory-utilization 0.90` to `0.95` to maximize KV cache capacity.
-- For image-only workloads (no audio), pass `--limit-mm-per-prompt audio=0` to skip audio encoder memory allocation.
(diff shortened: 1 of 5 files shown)


esmeetu:
Good Catch! Can we use `--limit-mm-per-prompt '{"image": 4, "audio": 1}'`, which looks more clean.

Aratako:
Thanks, updated the multi-modality examples to use the JSON form. I kept single-modality overrides in dotted form for readability, but I’m happy to switch those to JSON too if you prefer full consistency.

esmeetu:
@Aratako LGTM! Thanks for your contribution.

PR #463, opened 2026-05-14
update Gemma4 recipe for Intel cpu

louie-tsai:
Update Gemma4 for Intel Xeon 6 CPU

### DIFF
--- a/Google/Gemma4.md
+++ b/Google/Gemma4.md
@@ -10,15 +10,15 @@ Gemma 4 models are supported on NVIDIA GPUs, AMD GPUs, Google Cloud TPUs and Int
 
 | Model | Parameters | Min NVIDIA GPUs (BF16) | Min AMD GPUs (BF16) | Min TPUs | Min Xeon 6 CPUs |HuggingFace |
 |-------|-----------|------------------------|---------------------|----------|--------|-------------|
-| Gemma 4 E2B IT | effective 2B | 1× (24 GB+) | 1× MI300X/MI325X/MI350X/MI355X | - | 1x NUMA node | [google/gemma-4-E2B-it](https://huggingface.co/google/gemma-4-E2B-it) |
-| Gemma 4 E4B IT | effective 4B | 1× (24 GB+) | 1× MI300X/MI325X/MI350X/MI355X | - | 1x NUMA node | [google/gemma-4-E4B-it](https://huggingface.co/google/gemma-4-E4B-it) |
+| Gemma 4 E2B IT | effective 2B | 1× (24 GB+) | 1× MI300X/MI325X/MI350X/MI355X | - | 1×  NUMA node | [google/gemma-4-E2B-it](https://huggingface.co/google/gemma-4-E2B-it) |
+| Gemma 4 E4B IT | effective 4B | 1× (24 GB+) | 1× MI300X/MI325X/MI350X/MI355X | - | 1×  NUMA node | [google/gemma-4-E4B-it](https://huggingface.co/google/gemma-4-E4B-it) |
 | Gemma 4 31B IT | 31B | 1× (80 GB) | 1× MI300X/MI325X/MI350X/MI355X | 4× Trillium / 1× Ironwood | - | [google/gemma-4-31B-it](https://huggingface.co/google/gemma-4-31B-it) |
 
 ### Mixture-of-Experts (MoE) Models
 
 | Model | Total / Active Params | Min NVIDIA GPUs (BF16) | Min AMD GPUs (BF16) | Min TPUs | Min Xeon 6 CPUs | HuggingFace |
 |-------|----------------------|------------------------|---------------------|----------|----------|-------------|
-| Gemma 4 26B-A4B IT | 26B / 4B active | 1× (80 GB) | 1× MI300X/MI325X/MI350X/MI355X | 4× Trillium / 1× Ironwood | 2x NUMA node | [google/gemma-4-26B-A4B-it](https://huggingface.co/google/gemma-4-26B-A4B-it) |
+| Gemma 4 26B-A4B IT | 26B / 4B active | 1× (80 GB) | 1× MI300X/MI325X/MI350X/MI355X | 4× Trillium / 1× Ironwood | 2×  NUMA node | [google/gemma-4-26B-A4B-it](https://huggingface.co/google/gemma-4-26B-A4B-it) |
 
 ### Key Architecture Features
 
@@ -55,7 +55,7 @@ uv pip install vllm --pre \
 ```
 
 ### pip (Intel Xeon 6 CPUs)
-Please follow Intel/AMD x86 session in [CPU pre-build-wheels](https://docs.vllm.ai/en/latest/getting_started/installation/cpu/#pre-built-wheels)
+For Intel and AMD x86 CPUs, follow the [CPU pre-built wheels](https://docs.vllm.ai/en/latest/getting_started/installation/cpu/#pre-built-wheels) installation instructions.
 
 ### Docker
 
@@ -175,7 +175,7 @@ docker run -itd --name gemma4-rocm \
 
 ### Intel Xeon 6 Deployment via Docker
 
-Launch the CPU X86 vLLM Docker container where <MODEL> is your desired Google Gemma 4 model:
+Launch the x86 CPU vLLM Docker container, replacing `<MODEL>` with the desired Google Gemma 4 model:
 
 ```bash
 docker run -itd --name gemma4-cpu \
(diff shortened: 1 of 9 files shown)


bigPYJ1151:
Hi @louie-tsai looks the DCO check failed. https://github.com/vllm-project/recipes/pull/463/checks?check_run_id=76112906444

louie-tsai:
> Hi @louie-tsai looks the DCO check failed. https://github.com/vllm-project/recipes/pull/463/checks?check_run_id=76112906444

@bigPYJ1151  fixed it accordingly

models/Google/gemma-4-26B-A4B-it.yaml
```diff
@@ -97,11 +99,27 @@ compatible_strategies:
...
+      VLLM_RPC_TIMEOUT: "100000"
+    extra_args:
+      - "--dtype"
+      - "bfloat16"
+      - "--distributed-executor-backend"
+      - "mp"
+      - "--max-num-batched-tokens"
+      - "2048"
+      - "--max-num-seqs"
+      - "256"
```
bigPYJ1151:
```suggestion
hardware_overrides:
  cpu:
    extra_env:
      VLLM_CPU_KVCACHE_SPACE: "40"
      VLLM_CPU_ATTN_SPLIT_KV: "0"
```

Simplified settings, and Gemma is natively BF16. Same for others

Google/Gemma4.md
```diff
@@ -169,6 +173,23 @@ docker run -itd --name gemma4-rocm \
...
 ```
 
+### Intel Xeon 6 Deployment via Docker
+
+Launch the x86 CPU vLLM Docker container, replacing `<MODEL>` with the desired Google Gemma 4 model:
+
+```bash
+docker run -itd --name gemma4-cpu \
+    --network=host \
+    --shm-size 16G \
```
bigPYJ1151, line 183: Usually 2-4G is enough.

models/Google/gemma-4-26B-A4B-it.yaml
```diff
@@ -97,11 +99,27 @@ compatible_strategies:
...
 
 strategy_overrides:
   single_node_tp:
-    tp: 1
+    tp: 2
```
bigPYJ1151: This will change for all hardware.

models/Google/gemma-4-26B-A4B-it.yaml
```diff
@@ -298,3 +346,4 @@ guide: |
   - [vLLM Gemma 4 tool-call template](https://github.com/vllm-project/vllm/blob/main/examples/tool_chat_template_gemma4.jinja)
   - [TPU recipes: Trillium](https://github.com/AI-Hypercomputer/tpu-recipes/tree/main/inference/trillium/vLLM/Gemma4)
   - [TPU recipes: Ironwood](https://github.com/AI-Hypercomputer/tpu-recipes/blob/main/inference/ironwood/vLLM/Gemma4/)
+  - [ Xeon 6 recipe: Gemma 4 26B-A4B IT](https://aiswcatalog.intel.com/models/google-gemma-4-26b-a4b-it)
```
bigPYJ1151:
```suggestion
  - [Xeon 6 recipe: Gemma 4 26B-A4B IT](https://aiswcatalog.intel.com/models/google-gemma-4-26b-a4b-it)
```

models/Google/gemma-4-E4B-it.yaml
```diff
@@ -271,3 +319,4 @@ guide: |
   - [vLLM Gemma 4 tool-call template](https://github.com/vllm-project/vllm/blob/main/examples/tool_chat_template_gemma4.jinja)
   - [TPU recipes: Trillium](https://github.com/AI-Hypercomputer/tpu-recipes/tree/main/inference/trillium/vLLM/Gemma4)
   - [TPU recipes: Ironwood](https://github.com/AI-Hypercomputer/tpu-recipes/blob/main/inference/ironwood/vLLM/Gemma4/)
+  - [ Xeon 6 recipe: Gemma 4 E4B IT](https://aiswcatalog.intel.com/models/google-gemma-4-e4b-it)
```
bigPYJ1151:
```suggestion
  - [Xeon 6 recipe: Gemma 4 E4B IT](https://aiswcatalog.intel.com/models/google-gemma-4-e4b-it)
```

src/lib/command-synthesis.js
```diff
@@ -264,8 +266,10 @@ export function computeDockerMeta(recipe, variant, hwProfile) {
     ? "--privileged --network host \\\n  -v /dev/shm:/dev/shm"
     : isAmd
       ? "--device=/dev/kfd --device=/dev/dri \\\n  --security-opt seccomp=unconfined --group-add video"
+    : isIntel
+      ? "--shm-size=16g"
```
bigPYJ1151, line 270: Same, it is too large.

src/lib/command-synthesis.js
```diff
@@ -280,6 +284,9 @@ function dockerGpuArgv(meta) {
       "--group-add", "video",
     ];
   }
+  if (meta.isIntel) {
+    return ["--shm-size", "16g"];
```
bigPYJ1151, line 288: Same

src/lib/command-synthesis.js
bigPYJ1151: BTW, XPU is also going to be added, perhaps we need to distinguish CPU and XPU here.

taxonomy.yaml
```diff
@@ -112,6 +112,27 @@ hardware_profiles:
...
+
+  xeon5:
+    brand: Intel
+    generation: cpu
+    display_name: "Xeon 5"
+    description: "Intel Xeon 5 CPU node · x86 CPU backend"
+    gpu_count: 1
+    vram_gb: 512
+    multi_node: false
+    restricted: true
```
bigPYJ1151, line 134: Perhaps we can add more device count here, such as ```4```. The recipe will calculate a available count interval based on each model size.

For example, if set the ```gpu_count=4, vram_gb=512GB```, the device will be displayed as ```Xeon6 4x128G```. Then for the 26B model (64GB) the min count is 1, available interval is ```[1,4]```. The bottom of hardware block will display ```This recipe runs on 1 of 4 GPUs on the selected node — add`--tensor-parallel-size N` via Advanced to scale up```

models/Google/gemma-4-26B-A4B-it.yaml
louie-tsai: removed most of the args and envs accordingly.
still keep distributed-executor-backend = mp since cpu works good with mp.

Google/Gemma4.md
louie-tsai: since other hardware architectures also use 16G, so also keep the same value to align with other arches.

models/Google/gemma-4-26B-A4B-it.yaml
louie-tsai: changed it accordingly

src/lib/command-synthesis.js
louie-tsai: since other hardware architectures also use 16G, so also keep the same value to align with other arches.

louie-tsai: for XPU, what platforms we want to cover? maybe we create a seperate PR for XPU?

src/lib/command-synthesis.js
louie-tsai: since other hardware architectures also use 16G, so also keep the same value to align with other arches.

taxonomy.yaml
louie-tsai: Xeon6 4x128G looks confusing to me. it looks like a 4 nodes cases to me..  also we don't know how many numa nodes on users' Xeon system, so it might be hard to refer 4 as number of NUMA nodes for all cases...

models/Google/gemma-4-26B-A4B-it.yaml
bigPYJ1151: It will always be set internally after the new thread binding method.

src/lib/command-synthesis.js
bigPYJ1151: Yes, we don't need add XPU in this PR. I am just thinking about whether to reserve a label for it. Because ```isIntel``` is only used for CPU for now, and XPU has different container arguments.

taxonomy.yaml
bigPYJ1151: I think here should provide a top configuration of a single instance. Like NV GPU configurations include 8 GPUs. Users should adjust according to their system. For GNR, at most will include 2x3 NUMA nodes, but I don't prefer to highlight uneven number so just suggest 2x2 configuration...

models/Google/gemma-4-26B-A4B-it.yaml
louie-tsai: removed it accordingly

src/lib/command-synthesis.js
louie-tsai: Current codes don't cover hardware variants under same brand.  once we need XPU, we might need to change Flags from brandKey to hwKey.   Right now, they use brandKey for all docker images and arguements...

taxonomy.yaml
louie-tsai: ok. I added " CPU NUMA node " into description and changed according to your input.  we also have 4 sockets and 8 NUMA nodes GNR, but it should be a corner case.

louie-tsai:
@bigPYJ1151  please help to check changes for your input

src/lib/command-synthesis.js
bigPYJ1151: Ok... Let's leave this problem when add XPU recipes

bigPYJ1151 (approved):

PR #474, opened 2026-05-18
Update MiniMax M2.5 H200 recipe

anish-shanbhag:
Summary: Mark MiniMax-M2.5 verified on H200 and align the recipe with SemiAnalysisAI/InferenceX#1354. Pin vLLM image/min version to v0.20.2 and add the FP8 KV cache, FlashInfer attention/autotune, Triton MoE, and MiniMax QK norm fusion settings.


esmeetu:
LGTM. Thanks!

PR #480, opened 2026-05-20
Add DGX Spark Qwen3.6 recipe

Kh4L:
## Summary
- Add restricted DGX Spark GB10 hardware profile and picker support
- Add a DGX Spark recipe for nvidia/Qwen3.6-35B-A3B-NVFP4
- Keep default_hardware in the JSON API so the interactive command builder respects the Spark default

## Testing
- ruby -ryaml parse for all recipe, strategy, and taxonomy YAML files
- command synthesis smoke test for DGX Spark default hardware, Qwen ModelOpt args, env vars, and default MTP config


esmeetu:
I would like put this model like this https://recipes.vllm.ai/moonshotai/Kimi-K2.5?variant=nvfp4.

PR #483, opened 2026-05-21
Remove --no-async-scheduling from pd_cluster strategy

### LINKED ISSUE #479: Why we need `no-async-scheduling` in P/D disaggregation
It seems NixlConnector doesn't conflict with --no-async-scheduling

but only conflict with `--distributed-executor-backend ray`

Image

`RayDistributedExecutor` does not support async scheduling because its Ray Compiled DAG is always built with `enable_asyncio=False`
 ([ray_executor.py#L459]

`--distributed-executor-backend ray` or implicitly when running inside a Ray placement group — may silently disable `async scheduling`

https://recipes.vllm.ai/deepseek-ai/DeepSeek-V4-Flash?hardware=h100&features=tool_calling%2Creasoning%2Cspec_decoding&strategy=pd_cluster&prefill_nodes=2&prefill_rank=1

panpan0000:
## Summary

Remove `--no-async-scheduling` from `strategies/pd_cluster.yaml` prefill args.

RayExecutorV2 (the default since vLLM 0.11.x) supports async scheduling, so this flag is no longer needed and was causing a performance regression for users on the default executor.

Closes #479

## Changes

- `strategies/pd_cluster.yaml`: removed `--no-async-scheduling` from prefill `vllm_args`

## Test plan

- [x] `node scripts/build-recipes-api.mjs` — JSON API builds cleanly
- [x] PD cluster commands no longer include the unnecessary `--no-async-scheduling` flag

### DIFF
--- a/strategies/pd_cluster.yaml
+++ b/strategies/pd_cluster.yaml
@@ -23,7 +23,6 @@ prefill:
     - "--kv-transfer-config"
     - '{"kv_connector":"NixlConnector","kv_role":"kv_producer","kv_load_failure_policy":"fail"}'
     - "--enforce-eager"
-    - "--no-async-scheduling"
   env:
     VLLM_USE_NCCL_SYMM_MEM: "1"
     NCCL_CUMEM_ENABLE: "1"


esmeetu:
LGTM! Thanks!

PR #484, opened 2026-05-21
Update Minimax2.5 H100

faradawn:
Update the MiniMax-M2.5 recipe to match the new H100 FP8 launch flags validated in SemiAnalysisAI/InferenceX#1516: drop the explicit cudagraph_mode from the compilation-config, switch the reasoning parser to minimax_m2_append_think, and document TEP=8 (tensor-parallel-size 8 with --enable-expert-parallel) as the recommended single-node H100 strategy in place of TP4+EP.

### DIFF
--- a/models/MiniMaxAI/MiniMax-M2.5.yaml
+++ b/models/MiniMaxAI/MiniMax-M2.5.yaml
@@ -195,6 +195,17 @@ guide: |
     --enable-auto-tool-choice
   ```
 
+  ### H100 NVFP4 (TP=8)
+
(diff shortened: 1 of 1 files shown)


faradawn:
NVFP4 H100 configs from https://github.com/SemiAnalysisAI/InferenceX/pull/1517/changes

esmeetu:
Could you clarify why we need to change the reasoning parser?

faradawn:
I think its to add support for thinking.

PR #510, opened 2026-06-04
Add Trillium and Ironwood configurations for Qwens, Llama, and GPT-OSS

mtsokol:


mtsokol:
Hi @esmeetu,
This is a follow-up after https://github.com/vllm-project/recipes/pull/477. It adds four new model pages (Qwens) that we support for Trillium/Ironwood, and includes last missing TPU items from:
https://github.com/AI-Hypercomputer/tpu-recipes/tree/main/inference

esmeetu:
As Qwen3/Qwen2.5-VL are not latest models, I think we might not need to add these models. Are you still using those?

mtsokol:
> As Qwen3/Qwen2.5-VL are not latest models, I think we might not need to add these models. Are you still using those?

Yes, these are still relevant for us - if you have strong preference to not add them I can remove it from the PR. Otherwise I'd prefer to land all of them.

esmeetu:
I'm fine with adding this. Thanks for your contribution!

mtsokol:
Awesome - thank you!

PR #513, opened 2026-06-05
chore: remove SAFETENSORS_FAST_GPU=1 env var

faradawn:
This flag is not required for functionality, thus I propose to remove them from the recipes. 

Part of the cleanup effort. https://github.com/vllm-project/recipes/issues/366


esmeetu:
@hongxiayang FYI, I think this env doesn't have affect on latest version.

hongxiayang:
> @hongxiayang FYI, I think this env doesn't have affect on latest version.

sure. thanks for letting me know.

PR #518, opened 2026-06-09
add Xeon 6 CPUs support for Meta models

louie-tsai:
Add Xeon 6 CPU support


models/meta-llama/Llama-3.1-8B-Instruct.yaml
```diff
@@ -65,6 +66,9 @@ hardware_overrides:
     extra_args: []
     extra_env:
       VLLM_ROCM_USE_AITER: "1"
+  cpu:
+    extra_env:
+      VLLM_CPU_KVCACHE_SPACE: "40"
```
bigPYJ1151: Same here, we can start to suggest to use the auto setting

louie-tsai: sure. changed it accordingly.

PR #519, opened 2026-06-09
add Xeon 6 support for Qwen models

louie-tsai:
add Xeon 6 support for Qwen models

### DIFF
--- a/Qwen/Qwen3.5.md
+++ b/Qwen/Qwen3.5.md
@@ -33,6 +33,7 @@ For Intel and AMD x86 CPUs, follow the [CPU pre-built wheels](https://docs.vllm.
 ```bash
 uv venv
 source .venv/bin/activate
+export VLLM_CPU_KVCACHE_SPACE=40 \
 export VLLM_VERSION=$(curl -s https://api.github.com/repos/vllm-project/vllm/releases/latest | jq -r .tag_name | sed 's/^v//')
 uv pip install https://github.com/vllm-project/vllm/releases/download/v${VLLM_VERSION}/vllm-${VLLM_VERSION}+cpu-cp38-abi3-manylinux_2_35_x86_64.whl --torch-backend cpu
 ```
--- a/Qwen/Qwen3.5.md
+++ b/Qwen/Qwen3.5.md
@@ -33,7 +33,6 @@ For Intel and AMD x86 CPUs, follow the [CPU pre-built wheels](https://docs.vllm.
 ```bash
 uv venv
 source .venv/bin/activate
-export VLLM_CPU_KVCACHE_SPACE=40 \
 export VLLM_VERSION=$(curl -s https://api.github.com/repos/vllm-project/vllm/releases/latest | jq -r .tag_name | sed 's/^v//')
 uv pip install https://github.com/vllm-project/vllm/releases/download/v${VLLM_VERSION}/vllm-${VLLM_VERSION}+cpu-cp38-abi3-manylinux_2_35_x86_64.whl --torch-backend cpu
 ```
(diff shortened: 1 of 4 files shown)


models/Qwen/Qwen3-4B.yaml
```diff
@@ -53,14 +54,17 @@ compatible_strategies:
   - single_node_tp
   - multi_node_tp
 
-hardware_overrides: {}
+hardware_overrides:
+  cpu:
+    extra_env:
+      VLLM_CPU_KVCACHE_SPACE: "40"
```
bigPYJ1151: Do you think we should start to encourage use auto KV cache setting? 😂

Qwen/Qwen3.5.md
```diff
@@ -28,6 +28,15 @@ source .venv/bin/activate
 uv pip install vllm --extra-index-url https://wheels.vllm.ai/rocm
 ```
 
+#### CPU
+For Intel and AMD x86 CPUs, follow the [CPU pre-built wheels](https://docs.vllm.ai/en/latest/getting_started/installation/cpu/#pre-built-wheels) installation instructions.
```
bigPYJ1151, line 32: We can emphasis the environment variable settings here. It is easily to be dismissed when use wheel.

models/Qwen/Qwen3-4B.yaml
louie-tsai: sounds good. do you mean adding "--enable-prefix-caching" as server parameter?

Qwen/Qwen3.5.md
louie-tsai: added it accordingly

models/Qwen/Qwen3-4B.yaml
bigPYJ1151: Not it. Currently without setting ```VLLM_CPU_KVCACHE_SPACE```, workers will profile available memory size on corresponding node for KV cache. And users can restrict total memory usage via ```--gpu-memory-utilization```, just like what GPU worker does.

louie-tsai: sounds great. any document for reference? I can test it and then change the recipe accordingly.

bigPYJ1151: No need of special arguments, just removing the ```VLLM_CPU_KVCACHE_SPACE``` can use auto memory size profiling. Each worker will firstly calculate total memory size to allocate via ```total_allocated_mem_size = total_memory_node_size x gpu_memory_utilization (0.95 by default)```, then do the warmup and compilation. Finally it will check ```consumed_memory``` of the worker process, and get available KV cache size by ```available_kv_cache_size = total_allocated_mem_size  - consumed_memory```.

ekintel:
Since Qwen3.5 is multimodal, the vLLM server seems to require an extra flag like `--limit-mm-per-prompt '{"image":0,"video":0}'` or `--language-model-only`. Are we okay incorporating one of those in the recipe? Also, we can use one of the configurations `--reasoning-parser qwen3` or `--reasoning-parser qwen3 --default-chat-template-kwargs '{"enable_thinking": false}'` to handle reasoning.

models/Qwen/Qwen3-4B.yaml
louie-tsai: changed it accordingly.

louie-tsai: By the way, why we have one more CPU core for first rank?
image

louie-tsai:
> Qwen3.5 is multimodal, the vLLM server seems to require an extra flag like

@ekintel 
I don't see those parameters on other hardwares in current Qwen3.5 recipe.  
if we want to add multitmodal recipe, we might need to also add those parameters to all hardwares.
assuming to have another PR to cover that change. 
Hope that makes sense to you.

models/Qwen/Qwen3-4B.yaml
bigPYJ1151: Because previously the reserved CPU number is a per-rank number. Now it is a global number.

PR #535, opened 2026-06-12
[ROCm] Minimax-M3: add AMD command samples, 1M context length command sample and fp8-kv notes

hongxiayang:
Enhancing the Minimax-M3 recipe for Day 0
- Added AMD command samples, 
- 1M context length command sample 
- fp8-kv notes, 
- also replaced the docker placeholder (pending release)


hongxiayang:
cc @xianbaoqian

PR #554, opened 2026-06-16
Add vLLM-Omni TTS recipes (12 models across 8 providers)

linyueqian:
## Summary

Adds 12 structured YAML recipes (`tasks: [omni]`) for text-to-speech models served via **vLLM-Omni** through the OpenAI-compatible `/v1/audio/speech` API. This brings the TTS side of vLLM-Omni onto recipes.vllm.ai, complementing the existing `stabilityai/stable-audio-open-1.0` audio recipe and aligning with vllm-project/vllm-omni#2645.

## Recipes added

| Provider | Recipe |
|---|---|
| Qwen | `Qwen/Qwen3-TTS-12Hz-1.7B-CustomVoice` (VoiceDesign / Base / 0.6B as variants) |
| Mistral AI | `mistralai/Voxtral-4B-TTS-2603` |
| Fish Audio | `fishaudio/s2-pro` |
| OpenBMB | `openbmb/VoxCPM2` |
| OpenMOSS | `OpenMOSS-Team/MOSS-TTS`, `MOSS-TTS-Realtime`, `MOSS-TTSD-v1.0`, `MOSS-SoundEffect`, `MOSS-VoiceGenerator` |
| Z-AI | `zai-org/GLM-TTS` |
| Boson AI | `bosonai/higgs-audio-v3-tts-4b` |
| inclusionAI | `inclusionAI/Ming-omni-tts-0.5B` |

The five MOSS recipes follow OpenMOSS's official "MOSS-TTS Family" structure (five distinct production models) rather than collapsing them into one file.

## Provider registry

Adds `fishaudio`, `bosonai`, and `OpenMOSS-Team` to `src/lib/providers.js` (logos auto-fetched from HF avatars at build time).

## Notes on accuracy

- Facts are sourced from the upstream HF model cards/configs and the tested runbooks under `vllm-project/vllm-omni/recipes/`.
- Parameter counts come from HF safetensors metadata. `min_vllm_version` is set to each model's earliest known-good version (a floor, not the current release), so it does not need bumping as vLLM-Omni rebases.
- `omni.tasks` uses the `t2a` task with a per-task `endpoint: /v1/audio/speech` plus a custom curl override, so the rendered client command is correct for TTS rather than the diffusion `/v1/audio/generate` default.
- The four MossTTSDelay checkpoints (MOSS-TTS, TTSD, SoundEffect, VoiceGenerator) share `model_type=moss_tts_delay`, so their launch commands pass `--deploy-config` explicitly (auto-discovery cannot disambiguate them).

## Validation

- All 12 recipes render locally via `pnpm dev` (Next 15): every page returns HTTP 200 with no errors. Variant pills, the `/v1/audio/speech` client command, and provider logos all render.
- YAML parse and schema sanity checks pass for all 12 (model_id matches the file path, provider is registered, `omni.tasks` ids are valid, `variants.default` present).

cc @hsliuustc0106 @Sy0307 @gcanlin for review.


linyueqian:
Thanks for the review. Addressed in 60a85db:

- Added the `model` field to the Qwen3-TTS and `fishaudio/s2-pro` `/v1/audio/speech` curl examples.
- Switched the VoxCPM2 examples from the short alias `voxcpm2` to the served id `openbmb/VoxCPM2`.

One clarification for accuracy: vLLM-Omni's `/v1/audio/speech` handler treats `model` as part of the OpenAI schema but semantically ignores it (it always serves the single loaded model, see `serving_speech.py`), so the original examples were functional. The change is for consistency with the other recipes and with OpenAI SDK clients, which do require the `model` argument.

PR #559, opened 2026-06-17
update for glm52

JaredforReal:


models/zai-org/GLM-5.2.yaml
```diff
@@ -264,17 +261,6 @@ guide: |
 
   ## Troubleshooting
```
xianbaoqian, line 261: Do we want to mention here that if users ran into an issue with MTP, they need to switch to glm52 docker?

xianbaoqian (approved):

models/zai-org/GLM-5.2.yaml
JaredforReal: we can do that

models/zai-org/GLM-5.2.yaml
```diff
@@ -96,6 +94,7 @@ guide: |
     `main` branch.
   - **GPU:** 8xH200 or 8xH20 (141 GB each) for single-node FP8; 8xB200 (180 GB each) for the
     full 1M context.
+  - **DeepGEMM** (FP8): install via `install_deepgemm.sh` from the vLLM repo.
```
jeejeelee: We don't need to install DG with install_deepgemm.sh anymore.

JaredforReal: get, thanks

PR #561, opened 2026-06-17
[ROCm] GLM-5.2 recipe for MI300 and MI355

hongxiayang:
Add GLM-5.2 recipe for MI300 and MI355 for AMD/ROCm,

GSM8K 25-shot eval with MTP, num_concurrent=32:
 
flexible-extract exact_match: 0.9356 ± 0.0068
strict-match exact_match: 0.9363 ± 0.0067
 
MTP stats:
 
Draft steps: 45,009
Draft tokens: 225,045
Accepted tokens: 89,759
Acceptance per drafted token: 39.9%
Accepted tokens per draft step: 1.99 / 5
Accepted by position:
pos0: 42,166
pos1: 30,037
pos2: 12,771
pos3: 3,723
pos4: 1,062

### DIFF
--- a/models/zai-org/GLM-5.2.yaml
+++ b/models/zai-org/GLM-5.2.yaml
@@ -14,12 +14,15 @@ meta:
   hardware:
     b300: verified
     b200: verified
+    mi300x: verified
+    mi355x: verified
 
 model:
   model_id: "zai-org/GLM-5.2"
   min_vllm_version: "0.23.0"
   docker_image:
     nvidia: "vllm/vllm-openai:v0.23.0"
+    amd: "vllm/vllm-openai-rocm:nightly"
   architecture: moe
   parameter_count: "743B"
   active_parameters: "39B"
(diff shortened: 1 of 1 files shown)


models/zai-org/GLM-5.2.yaml
```diff
@@ -14,12 +14,15 @@ meta:
...
     b200: verified
+    mi300x: verified
+    mi355x: verified
 
 model:
   model_id: "zai-org/GLM-5.2"
   min_vllm_version: "0.23.0"
   docker_image:
     nvidia: "vllm/vllm-openai:v0.23.0"
+    amd: "vllm/vllm-openai-rocm:nightly"
```
xianbaoqian: Wondering if you know vllm/vllm-openai-rocm:v0.23.0  works? we would prefer to have a named version over nightly which changes over time.

xianbaoqian (commented):
Looking good. Thanks!

models/zai-org/GLM-5.2.yaml
hongxiayang: we need to use vllm/vllm-openai-rocm:nightly-4c626633159887b0f2c962058c17c78f1434556d.
THe v.0.23.0 does not work

hongxiayang: @xianbaoqian : updated the docker image

tjtanaa: @xianbaoqian nightly image like on CUDA, only have 14-days retention period, it will be removed after 14 days. 

At the moment, for functionality, we can allow the use of nightly, when we have a stable release that works then only swap to the `latest` tag. The major accuracy bugfix for mi300x was just merged 2 days ago. So that's why nightly is needed.

PR #566, opened 2026-06-18
GLM-5.2: use fp8 KV cache on Hopper (Blackwell/AMD keep fp8_e4m3)

xianbaoqian:
Scope --kv-cache-dtype per GPU generation via hardware_overrides.hopper: H100/H200 render --kv-cache-dtype fp8, while Blackwell (B200/B300) and AMD keep base_args' fp8_e4m3 unchanged. The synthesizer's last-wins dedupe lets the Hopper override shadow the base flag, so the rendered command always carries a single --kv-cache-dtype (verified: h200=fp8, b200/mi300x=fp8_e4m3). Guide Docker + 8xH200 commands updated to match.

### DIFF
--- a/models/zai-org/GLM-5.2.yaml
+++ b/models/zai-org/GLM-5.2.yaml
@@ -76,6 +76,12 @@ compatible_strategies:
   - pd_cluster
 
 hardware_overrides:
+  hopper:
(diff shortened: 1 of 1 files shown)


xianbaoqian:
confirmed offline with z.ai team.

PR #567, opened 2026-06-18
fix(deepseek-v4-pro): cap multi_node_tep TP to 8 for FP8 weights

### LINKED ISSUE #497: [Bug] DeepSeek-V4-Pro multi_node_tep renders invalid TP=16 command for FP8 weights
### Problem

The current `DeepSeek-V4-Pro` recipe renders an invalid `multi_node_tep` command for a 2-node H200 deployment.

On the recipes site, selecting:

- model: `deepseek-ai/DeepSeek-V4-Pro`
- hardware: `H200`
- nodes: `2`
- strategy: `multi_node_tep`

renders a command with:

```bash
--tensor-parallel-size 16
```

For the default FP8 checkpoint, this is not a valid configuration.

### Why this is wrong

DeepSeek V4 Pro uses `moe_intermediate_size = 3072`, and the default FP8 weight format requires the partitioned input dimension to be compatible with the FP8 block size.

With `TP=16`:

- `3072 / 16 = 192`

This fails for the default FP8 weights. The related upstream vLLM issue is:

- https://github.com/vllm-project/vllm/issues/40955

There is also an upstream PR addressing cross-node TP=16 FP8 serving:

- https://github.com/vllm-project/vllm/pull/41312

Until that lands and is available in a released vLLM version, the recipe should not render `TP=16` as the default multi-node TEP configuration for this model.

### Suggested fix

For `DeepSeek-V4-Pro`, the `multi_node_tep` recipe should be overridden to use:

```bash
--tensor-parallel-size 8
```

for the current default FP8 weights, instead of deriving TP from total GPUs across nodes.

mahadrehmann:
## Problem
The DeepSeek-V4-Pro recipe renders `--tensor-parallel-size 16` for a 
2-node H200 `multi_node_tep` deployment. For FP8 weights, 
`moe_intermediate_size = 3072` and `3072 / 16 = 192`, which is not a 
multiple of the FP8 block size (128), breaking block-quantized GEMMs 
at runtime.

## Fix
Added `--tensor-parallel-size 8` to `strategy_overrides.multi_node_tep.extra_args`.
Via `dedupeArgs` last-wins logic in `command-synthesis.js`, this overrides 
the auto-computed TP=16.

## Test Plan
- YAML validated: `python3 -c "import yaml; yaml.safe_load(open('models/deepseek-ai/DeepSeek-V4-Pro.yaml')); print('YAML parses OK')"` → YAML parses OK
- Logic verified: `--tensor-parallel-size 8` in `extra_args` (step 4) 
  overrides auto-computed TP=16 (step 3) via `dedupeArgs` last-wins in `command-synthesis.js`
- Math: `3072 / 8 = 384` ✓ multiple of FP8 block size (128)

## Note
NVFP4 on `multi_node_tep` also gets TP=8 — valid though suboptimal. 
Per-variant strategy overrides would require changes to `command-synthesis.js`, 
out of scope here.

## Revert condition
Remove this override once `vllm-project/vllm#41312` ships in a stable release.

Fixes #497
Ref: vllm-project/vllm#40955
Ref: vllm-project/vllm#41312

### DIFF
--- a/models/deepseek-ai/DeepSeek-V4-Pro.yaml
+++ b/models/deepseek-ai/DeepSeek-V4-Pro.yaml
@@ -146,7 +146,7 @@ strategy_overrides:
       - "--compilation-config"
       - '{"cudagraph_mode":"FULL_AND_PIECEWISE", "custom_ops":["all"]}'
   multi_node_tep:
-    # moe_intermediate_size=3072 / TP must divide evenly into FP8 block_n=128.
+    # moe_intermediate_size=3072 / TP must be a multiple of FP8 block_n=128.
     # TP=16 (2 × 8-GPU nodes) gives 3072/16=192, which is not a multiple of 128
     # and breaks FP8 block-quantized GEMMs (vllm-project/vllm#40955). Lock TP=8
     # until vllm-project/vllm#41312 ships in a stable release; revert this then.
--- a/models/deepseek-ai/DeepSeek-V4-Pro.yaml
+++ b/models/deepseek-ai/DeepSeek-V4-Pro.yaml
@@ -146,7 +146,13 @@ strategy_overrides:
       - "--compilation-config"
       - '{"cudagraph_mode":"FULL_AND_PIECEWISE", "custom_ops":["all"]}'
   multi_node_tep:
+    # moe_intermediate_size=3072 / TP must divide evenly into FP8 block_n=128.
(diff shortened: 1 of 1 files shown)


majunze2001:
[#46001](https://github.com/vllm-project/vllm/pull/46001) has been merged

mahadrehmann:
Thanks for the heads up @majunze2001.  Closing since #46001 covers this

PR #575, opened 2026-06-24
Add LiquidAI LFM2.5 recipes (dense, MoE, VL)

vincentzed:
Adds the first **LiquidAI** provider entry, covering the LFM2.5 language and
vision-language model family. All recipes serve via vLLM's native LFM2
architectures — no `--trust-remote-code` needed.

### Recipes added (`models/LiquidAI/`)

| Recipe | Arch | Params | Features |
|--------|------|--------|----------|
| LFM2.5-350M | dense (`Lfm2ForCausalLM`) | 350M | tools |
| LFM2.5-1.2B-Instruct | dense | 1.2B | tools |
| LFM2.5-1.2B-Thinking | dense | 1.2B | reasoning + tools |
| LFM2.5-1.2B-JP | dense | 1.2B | — |
| LFM2.5-1.2B-JP-202606 | dense | 1.2B | tools |
| LFM2.5-1.2B-Base | dense | 1.2B | completions (base) |
| LFM2.5-8B-A1B | MoE (`Lfm2MoeForCausalLM`) | 8B / ~1B active | reasoning + tools |
| LFM2.5-VL-450M | VL (`Lfm2VlForConditionalGeneration`) | 450M | image+text |
| LFM2.5-VL-1.6B | VL | 1.6B | image+text |

Plus a `LiquidAI/LFM2.5.md` provider guide (install, supported-models table,
serve commands, reasoning/tools/vision usage, per-card recommended sampling)
and a `LiquidAI` entry in `src/lib/providers.js`.

### Notes

- **vLLM ≥ 0.23.0** — all three architectures (dense / MoE / VL) and the
  `lfm2` tool parser ship in the 0.23.0 stable release.
- **Tool calling**: `--enable-auto-tool-choice --tool-call-parser lfm2`
  (vLLM's native `Lfm2ToolParser` for LFM2/LFM2.5 Pythonic tool calls).
- **Reasoning** (8B-A1B, 1.2B-Thinking): `--reasoning-parser qwen3` — the
  correct parser for LFM2.5's `<think>…</think>` format. The models open the
  `<think>` channel for harder reasoning turns (a simple prompt may be answered
  inline), and the parser separates it into `reasoning_content` when present.
- **Sampling** is documented per model card in the guides (not baked into the
  serve command).

### Validation

Every recipe's exact generated serve command was run on **NVIDIA H100** with
`vllm/vllm-openai:v0.23.0` (one container per model): all 9 reach `/health`,
[...]

PR #578, opened 2026-06-25
LFM2.5: fix context_length to 32K (per model cards) + Blackwell oink + copy cleanup

vincentzed:
Follow-up to #575 (LFM2.5 recipes) — a context-length correction, a measured perf flag, and copy cleanup.

### Fix `context_length` to 32K (per the model cards)

#575 set `context_length: 128000` on every recipe (config.json's `max_position_embeddings`, which
is inherited RoPE config), but the model cards document **32,768** for the whole family **except
8B-A1B (128K)**. This corrects the field and the "128K" copy on the 8 affected recipes and the
guide's tables. (8B-A1B stays 128K; `LiquidAI/LFM2.5-230M` was added separately in #576 with the
correct 32K and is left untouched here.)

### Performance: oink on Blackwell

`VLLM_USE_OINK_OPS=1` routes RMSNorm to the `oink` kernels bundled in the `vllm-openai` image on
Blackwell. **~+2.6%** output throughput on **B200** (1.2B-Instruct, vLLM 0.23.0), output identical.
Applied automatically via `hardware_overrides.blackwell` (native fallback elsewhere).

### Speed-of-Light Tuning section + copy cleanup

- Adds a `LiquidAI/LFM2.5.md` section: `-O3` (~+2%, opt-in, trades startup), `VLLM_USE_FASTOKENS=1` (TTFT win on tokenization-bound loads); other knobs make little difference (measured with `vllm bench serve`).
- Drops self-referential boilerplate: "no `--trust-remote-code` required", "Verified on H100", "validated end-to-end".

`node scripts/build-recipes-api.mjs` passes; MkDocs `--strict` passes.

### DIFF
--- a/LiquidAI/LFM2.5.md
+++ b/LiquidAI/LFM2.5.md
@@ -4,17 +4,14 @@
 efficient open-weight models built on the **LFM2 hybrid backbone** — short-range gated
 convolution blocks interleaved with grouped-query attention. The hybrid design keeps the KV
 cache small and decode fast, so LFM2.5 models punch above their weight while remaining cheap to
-serve, down to edge / on-device GPUs. The family spans dense chat models (350M, 1.2B), a
+serve, down to edge / on-device GPUs. The family spans dense chat models (230M, 350M, 1.2B), a
 mixture-of-experts model (8B-A1B), reasoning and Japanese variants, base checkpoints, and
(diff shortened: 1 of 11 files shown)


vincentzed:
@ywang96 and @esmeetu Thank you in advance!

PR #592, opened 2026-06-29
Update ERNIE 4.5 21B ROCm serve command

haic0:
## Summary
- Add verified hardware tags for ERNIE 4.5 21B, including H200 and AMD MI300X / MI325X / MI355X.
- Keep AMD ROCm env and GPU memory guidance while avoiding duplicate `--tensor-parallel-size` injection from hardware overrides.
- Use `strategy_overrides.single_node_tp.tp: 4` as the single source for the TP=4 setting.

## Test plan
- Parsed `models/baidu/ERNIE-4.5-21B-A3B-PT.yaml` with PyYAML and checked top-level key order.
- Verified AMD hardware tags and confirmed `hardware_overrides.amd.extra_args` no longer contains `--tensor-parallel-size`.
- Ran `git diff --check -- models/baidu/ERNIE-4.5-21B-A3B-PT.yaml`.
- Ran `node scripts/build-recipes-api.mjs`.

### DIFF
--- a/models/baidu/ERNIE-4.5-21B-A3B-PT.yaml
+++ b/models/baidu/ERNIE-4.5-21B-A3B-PT.yaml
@@ -17,7 +17,8 @@ model:
   parameter_count: "21B"
   active_parameters: "3B"
   context_length: 131072
-  base_args: []
+  base_args:
+    - "--trust-remote-code"
   base_env: {}
 
 features:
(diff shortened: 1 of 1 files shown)


models/baidu/ERNIE-4.5-21B-A3B-PT.yaml
```diff
@@ -18,7 +18,8 @@ model:
   parameter_count: "21B"
   active_parameters: "3B"
   context_length: 131072
-  base_args: []
+  base_args:
+    - "--trust-remote-code"
   base_env: {}
```
tjtanaa, line 29: Add hardware verified tag.

haic0:
@tjtanaa  Added, pls check, thanks!

models/baidu/ERNIE-4.5-21B-A3B-PT.yaml
```diff
@@ -97,10 +105,13 @@ guide: |
 
   ## Launch commands
 
-  21B on 1x80GB GPU:
+  21B on AMD ROCm with 4-way tensor parallelism:
```
tjtanaa: @haic0 you modified the cuda command, can you introduce the commands somewhere that make sense? Thanks.

PR #594, opened 2026-06-29
Update Ling 2.6 Flash ROCm command

haic0:
## Summary
- Adjust Ling 2.6 Flash to the provided TP=2 ROCm launch command.
- Aligns the recipe launch guidance with the provided vLLM serve command.

## Test plan
- Ran `node scripts/build-recipes-api.mjs` on the complete validated recipe update set.


models/inclusionAI/Ling-2.6-flash.yaml
```diff
@@ -78,7 +84,7 @@ guide: |
     --group-add video \
     -e VLLM_ROCM_USE_AITER=1 \
     vllm/vllm-openai-rocm:v0.20.2 \
-      inclusionAI/Ling-2.6-flash \
+      vllm serve inclusionAI/Ling-2.6-flash \
```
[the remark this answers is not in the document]
tjtanaa: we don't need to add `vllm serve` 

entrypoint of `vllm/vllm-openai-rocm:v0.20.2 ` image is already `vllm serve`

models/inclusionAI/Ling-2.6-flash.yaml
```diff
@@ -51,7 +49,7 @@ hardware_overrides:
 
 strategy_overrides:
```
tjtanaa, line 52: find a way to make it so that nvidia uses tp4 and amd uses tp2

tjtanaa: and remember to update the verified

PR #596, opened 2026-06-29
Update Laguna XS.2 serve command

haic0:
## Summary
- Adjust Laguna XS.2 to the provided trust-remote-code and TP=8 launch command.
- Aligns the recipe launch guidance with the provided vLLM serve command.

## Test plan
- Ran `node scripts/build-recipes-api.mjs` on the complete validated recipe update set.

### DIFF
--- a/models/poolside/Laguna-XS.2.yaml
+++ b/models/poolside/Laguna-XS.2.yaml
@@ -48,7 +48,7 @@ variants:
   default:
     precision: bf16
     vram_minimum_gb: 80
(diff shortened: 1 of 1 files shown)


dllehr-amd:
This looks like it'll change the entire Laguna recipe to use TP8.  Including the NVIDIA hardware. I'm not sure this is a change we want to make lightly.  Can you explain why you're trying to do this?

dllehr-amd (approved):
Sorry wrong PR

dllehr-amd (changes requested):
Can you explain why we're changing the default to TP8 for everyone?

PR #603, opened 2026-06-30
docs: document You.com as alternative browser backend for gpt-oss

tyler5673:
The gpt-oss reference library ships a `YouComBackend` and the MCP browser server (`browser_server.py`) defaults to `BROWSER_BACKEND=youcom`, but the vLLM gpt-oss recipes only documented the Exa backend.

This adds a brief "Browser backend" section to both the 20b and 120b recipe guides showing how to use You.com via the MCP browser server with `YDC_API_KEY`.

The demo tool server path (`--tool-server demo`) still uses Exa by default and is noted as such. The You.com path applies to the MCP browser server, which defaults to `BROWSER_BACKEND=youcom`.

The `mcp run` and `vllm serve` commands are in separate code blocks since both are long-running blocking processes that need separate terminals.


esmeetu:
Sorry. Thanks for your contribution. But we don't plan add third-party mcp server in recipe page.

PR #605, opened 2026-07-01
Add MiniMax M3 NVFP4 for B200

faradawn:
Add the NVFP4 (nvidia/MiniMax-M3-NVFP4) Blackwell variant to the MiniMax-M3 recipe, validated on B200 with EAGLE3 speculative decoding, matching the InferenceX benchmark: --block-size 128, --trust-remote-code, and VLLM_FLOAT32_MATMUL_PRECISION=high, reusing the existing spec-decoding and text-only feature toggles. Based on SemiAnalysisAI/InferenceX#1933.


esmeetu:
Can you pass the DCO check?

faradawn:
Fixed Doc Sign and merge conflicts!

PR #614, opened 2026-07-06
chore: remove VLLM_USE_FLASHINFER_MOE_FP4 flag

faradawn:
After v0.23.1rc0, the flag VLLM_USE_FLASHINFER_MOE_FP4 seems to be auto selected. See `vllm/model_executor/layers/fused_moe/oracle/nvfp4.py`, `select_nvfp4_moe_backend()`.

Users can now use `--moe-backend` to specify. 

Part of the clean up effort: https://github.com/vllm-project/recipes/issues/366

### DIFF
--- a/models/MiniMaxAI/MiniMax-M2.5.yaml
+++ b/models/MiniMaxAI/MiniMax-M2.5.yaml
@@ -18,7 +18,7 @@ meta:
 
 model:
   model_id: "MiniMaxAI/MiniMax-M2.5"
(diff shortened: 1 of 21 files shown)


xinli-sw:
should docker version also be bumped?

esmeetu:
I think we don't need to update version? the model minimal version is the first version vLLM support. I think just replace env with `--moe-backend` for specific models is good for me.

PR #617, opened 2026-07-07
fix  h20*8 deepseek-v4-pro-dspark bug

yiminghub2024:
fix  h20*8 deepseek-v4-pro-dspark bug
--max-num-seqs must set to 16

recep cookbook  512 can not be started successful 

it have been test many times ~ 

thanks~

### DIFF
--- a/models/deepseek-ai/DeepSeek-V4-Pro.yaml
+++ b/models/deepseek-ai/DeepSeek-V4-Pro.yaml
@@ -136,7 +136,7 @@ hardware_overrides:
       - "--gpu-memory-utilization"
       - "0.95"
       - "--max-num-seqs"
-      - "512"
+      - "16"
       - "--max-num-batched-tokens"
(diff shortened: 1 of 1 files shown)


esmeetu:
@yiminghub2024 Thanks for your PR! BTW, could you help test the proper value for "--max-num-batched-tokens" on H20?

yiminghub2024:
yes ,i have tested , best setting for h20*8
--max-num-batched-tokens 16384   & --max-model-len 200000  is the best 
more than this ,it can not be start successful ,  if use this , the kvcache size is 312K , it will be one 200K request or 64K*5 request . 
thanks   ,if --max-num-batched-tokens 32768 ,  the --max-model-len auto  result only less than 10k ,its no use for product env   @esmeetu  i have fix this both on this pr

esmeetu:
Thank you!

PR #625, opened 2026-07-11
Add KV cache offloading (LMCache) feature to DeepSeek-V4-Pro

ApostaC:
## Summary

- New opt-in **KV offloading (LMCache)** feature on DeepSeek-V4-Pro: adds `--kv-transfer-config` with `LMCacheMPConnector` for KV cache offloading to a shared CPU-DRAM pool
- The companion `lmcache server` command renders as a selectable **LMCache Server** tab in the command block (same tab pattern as the PD-disagg Prefill/Decode/Router block)
- Optional `lmcache==0.5.1` extra-install entry (dimmed, excluded from Copy-all) plus guide notes on pool sizing and port matching
- Gated to single-node strategies: on PD cluster the pill renders disabled, so the LMCache config never clobbers the NixlConnector wiring
- Launch config verified on 8× B200/B300 (SemiAnalysisAI/InferenceX#2153)

## New feature-schema fields (generic, reusable by other recipes)

- `strategies:` — per-strategy allowlist; synthesis skips the feature's args and companion on excluded strategies, the UI strikes the pill through
- `companion:` — `{ label, description?, command }` helper process rendered as an extra command tab and exported as `companions[]` in the JSON API
- `label:` — pill-text override (default stays the title-cased key)

## Validation

image

- `node scripts/build-recipes-api.mjs` → ✓ 146 models, 8 strategies
- Synthesis checks: exactly one `--kv-transfer-config` per command on single-node strategies; pd_cluster keeps NixlConnector; feature off emits nothing
- Verified end-to-end in the command builder UI (tab switching, strategy gating, URL sync)


esmeetu:
Closed as of #628. Thanks for your contribution!

PR #629, opened 2026-07-12
docs: add DeepSeek V4 Pro DEP prefill scheduling

xinli-sw:


yiminghub2024:
what's the benifit of config this ? can you give the data compare of add this ? include tps & latecy % concurrent increase ? @xinli-sw

PR #650, opened 2026-07-17
[AMD] add kimi-k2.5 mxfp4 recipe

hongxiayang:
Add recipe and notes on serving [`amd/Kimi-K2.5-MXFP4`](https://huggingface.co/amd/Kimi-K2.5-MXFP4)


hongxiayang:
cc @xianbaoqian

PR #652, opened 2026-07-17
Add GLM-5.2 low-latency serving recipe

simon-mo:
## What changed

- Pin the GLM-5.2 NVFP4 low-latency path to `vllm/vllm-openai:v0.25.1`.
- Configure the single-node Blackwell command with TP8, FP8 KV cache, MTP=5, the fused GLM model-class override, breakable piecewise CUDA graphs, tool/reasoning parsers, and the stable served model name.
- Make Docker the primary install path and replace the stale `:glm52` guide example with the release-tested copy-paste command.
- Preserve variant-forced `default_modes` when building the static JSON API.

## Why

The public GLM-5.2 guide still showed the older image and a generic NVFP4 command, while the current low-latency path requires vLLM 0.25.1 and explicit fused-model/CUDA-graph configuration.

The API generator also excluded opt-in features unconditionally. As a result, promoted variant endpoints dropped MTP even when a variant's `default_modes` explicitly enabled it. The generator now mirrors the interactive command builder.

## Impact

Selecting the NVFP4 variant on B200/B300, or consuming `/nvidia/GLM-5.2-NVFP4.json`, produces the same low-latency command with MTP=5 and the pinned Docker image.

## Validation

- Directly inspected the generated native and Docker commands in `public/nvidia/GLM-5.2-NVFP4.json`.
- `git diff --check`

`pnpm lint` was not available as a non-interactive check because this repository has no ESLint configuration and `next lint` opens the first-run configuration prompt.

### DIFF
--- a/models/zai-org/GLM-5.2.yaml
+++ b/models/zai-org/GLM-5.2.yaml
@@ -3,7 +3,7 @@ meta:
   slug: "glm-5.2"
   provider: "GLM (Z-AI)"
   description: "GLM-5.2 — frontier-scale MoE language model (~743B total parameters, 39B active) with up to 5-token MTP speculative decoding and thinking mode"
-  date_updated: 2026-07-12
+  date_updated: 2026-07-17
   difficulty: advanced
   tasks:
     - text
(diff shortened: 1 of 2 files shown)


gaby:
@simon-mo Fix the merge conflict

PR #665, opened 2026-07-22
Add MI325X recipe for Laguna S 2.1

akii96:
## Validation
- Hardware: 8× AMD MI325X (`gfx942`)
- ROCm: 7.2
- vLLM image: `vllm/vllm-openai-rocm:nightly-387189c42997b27e2c04b5d97ef8190ffa2bf909`
- Chat-completion smoke test passed.
- `node scripts/build-recipes-api.mjs` passed with 149 models.

### DIFF
--- a/models/poolside/Laguna-S-2.1.yaml
+++ b/models/poolside/Laguna-S-2.1.yaml
@@ -4,7 +4,7 @@ meta:
   provider: "Poolside"
   description: "Poolside's 118B total / 8B activated MoE coding model with mixed sliding-window + global attention, native interleaved reasoning, and 256K context — the larger sibling of Laguna XS-2.1, tuned for agentic coding."
   date_added: 2026-07-21
-  date_updated: 2026-07-21
+  date_updated: 2026-07-22
   difficulty: intermediate
   tasks:
     - text
@@ -14,6 +14,7 @@ meta:
     - "poolside/Laguna-M.1"
   hardware:
     h200: verified
+    mi325x: verified
 
 model:
   model_id: "poolside/Laguna-S-2.1"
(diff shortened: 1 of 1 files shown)


models/poolside/Laguna-S-2.1.yaml
```diff
@@ -159,6 +167,20 @@ guide: |
...
+  VLLM_ROCM_USE_AITER=1 \
+  VLLM_ROCM_USE_AITER_FUSION_SHARED_EXPERTS=1 \
+  VLLM_ROCM_QUICK_REDUCE_QUANTIZATION=INT4 \
+  vllm serve poolside/Laguna-S-2.1 \
+    --trust-remote-code \
+    --tensor-parallel-size 8 \
+    --enable-auto-tool-choice \
+    --tool-call-parser poolside_v1 \
+    --reasoning-parser poolside_v1 \
+    --default-chat-template-kwargs '{"enable_thinking":true}'
```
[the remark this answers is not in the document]
akii96: Thanks for catching the inconsistency. I added `--max-model-len 262144` to the MI325X command.

But for full context, 8× MI325X provides substantially more VRAM headroom than the NVIDIA configurations referenced here, and the engine successfully initialized using Laguna’s native 1M maximum. The 256K cap is therefore not required for MI325X capacity; it is included to keep the recipe consistent with its declared context length and the other launch examples.

models/poolside/Laguna-S-2.1.yaml
```diff
@@ -160,6 +176,40 @@ guide: |
     --reasoning-parser poolside_v1
   ```
 
+  ### AMD ROCm — 8× MI325X (BF16, TP8)
+  ```bash
+  VLLM_ROCM_USE_AITER=1 \
```
esmeetu: These command lines looks duplicate with auto-generated command line above. Can we clean it?

models/poolside/Laguna-S-2.1.yaml
```diff
@@ -54,13 +55,28 @@ variants:
     precision: bf16
     vram_minimum_gb: 282
     description: "BF16 weights (~235GB) — needs a multi-GPU node (≥2× H200/B200, or a full HGX node)"
+    hardware_overrides:
+      mi325x:
+        docker_image: "vllm/vllm-openai-rocm:nightly-387189c42997b27e2c04b5d97ef8190ffa2bf909"
```
esmeetu: Does latest image support it now? I think we can remove this line to make it more general.

akii96: Yeah latest images do support it! So dropped it now!

models/poolside/Laguna-S-2.1.yaml
akii96: Thanks for looking into this @esmeetu I fixed this now! My bad as I assumed just like the H200 recipe I was supposed to repeat the recipe in the canvas below.

PR #686, opened 2026-07-27
Sort latest recipes by catalog addition date

xianbaoqian:
PS: I'm not entirely sure if we want to do this..

Currently Kimi is not listed as the first one because their HF reop being created way too early.

-----

## Summary

- add immutable `meta.date_added` metadata to all 150 recipes, backfilled from each recipe first landing commit on `main` using UTC committer dates
- preserve the original lineage across the two verified recipe renames and avoid Git copy-similarity false positives
- sort the homepage Latest recipes section and empty search Latest list by catalog addition date instead of Hugging Face repository creation time
- keep `date_updated` for modification timestamps and document both fields in the schema and recipe-authoring guidance

This makes the ranking deterministic in a fresh checkout and correctly puts `moonshotai/Kimi-K3` first even though its Hugging Face repository existed before launch.

## Validation

- audited 150/150 `date_added` values against direct main-branch add history with zero mismatches
- verified all model diffs are one-line metadata additions
- `node --check src/lib/recipes.js`
- `node --check src/lib/recipe-utils.js`
- verified generated Kimi-K3 JSON preserves both dates as strings


BugenZhao:
Just added a commit to backfill the added dates semantically (by backtracking the repo history, instead of only checking the created date of the YAML file)

PR #702, opened 2026-07-29
feat: add Inferact/MiniMax-M3-EAGLE3-GQA as selectable spec_decoding mode

xinli-sw:
## Summary

- Converts the `spec_decoding` feature on MiniMax-M3 from a flat `args[]` to a `modes` map with two selectable Eagle3 draft heads:
  - **MHA** (default): `Inferact/MiniMax-M3-EAGLE3` — 64 KV heads, highest acceptance rate (~2.7–3.5 mean accepted tokens on MT-Bench)
  - **GQA**: `Inferact/MiniMax-M3-EAGLE3-GQA` — 4 KV heads, 16× smaller draft KV cache, slight acceptance-rate trade-off (~2.67); better for long-context or high-batch deployments
- Both modes carry per-platform `hardware_overrides` for the correct `attention_backend` (FLASH_ATTN on NVIDIA, TRITON_ATTN on AMD)
- Adds an **EAGLE3 Speculative Decoding** section to the guide with a comparison table
- Adds a GQA example command in the NVFP4 section
- Adds both Inferact HF model card links to References

## Test plan

- [x] Run `node scripts/build-recipes-api.mjs` — should print `✓ JSON API` with no errors
- [x] Open the MiniMax-M3 recipe page, enable Spec decoding — verify MHA/GQA segmented row appears
- [x] Switch to GQA mode and confirm `Inferact/MiniMax-M3-EAGLE3-GQA` appears in the command

### DIFF
--- a/models/MiniMaxAI/MiniMax-M3.yaml
+++ b/models/MiniMaxAI/MiniMax-M3.yaml
@@ -52,16 +52,21 @@ features:
       - "--reasoning-parser"
       - "minimax_m3"
   spec_decoding:
-    description: "Eagle3 speculative decoding with the Inferact MiniMax-M3 draft head — accelerates decoding."
-    args:
-      - "--speculative-config"
-      - '{"method": "eagle3", "model": "Inferact/MiniMax-M3-EAGLE3", "num_speculative_tokens": 3, "attention_backend": "FLASH_ATTN"}'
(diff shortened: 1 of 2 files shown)


faradawn (commented):
LGTM

PR #705, opened 2026-07-30
Add filesystem tier option, grouped under one Offloading pill

Etelis:
Stacked on #704 (first commit; review the last commit here).

Adds the multi-tier configuration of the [OffloadingConnector](https://docs.vllm.ai/en/stable/features/kv_offloading_usage) (`TieringOffloadingSpec` with a filesystem secondary tier), and a small `kv_offload_groups` taxonomy block: options sharing a `group` collapse into one pill on the KV Offload row, with a nested "Offload Tiers" sub-row picking the member — the same idiom as Mooncake's merged pill + Store Topology row. Members are labeled by tier (CPU · CPU + Filesystem), so the future obj/p2p secondary tiers join as sub-row entries rather than new pills.

Also adds per-option `env` support, used here to pin `PYTHONHASHSEED=0` — block content hashes are seeded from `NONE_HASH`, so without it the on-disk cache lands under different filenames every restart and never hits.

Both members are `requires_opt_in` and stay disabled on every recipe, as in #704.

**Testing**
- `node scripts/build-recipes-api.mjs` — 151 models, no connector leakage into per-model JSON.
- Gate suite: both members fail closed without opt-in; opt-in is per member; fs member emits the tiering config + `PYTHONHASHSEED=0`; brand/pd gates hold; Simple/Off/Mooncake unchanged.
- `pnpm dev` render of both states: merged pill disabled by default; with a locally opted-in recipe the Offload Tiers sub-row renders and the command carries the tiering config and env.


Etelis:
Consolidated into #704.

PR #708, opened 2026-07-30
Render one command per node in multi-node command blocks

esmeetu:
## Problem

Multi-node deployments only ever rendered **two** commands — head (rank 0) and one `worker` hard-pinned to rank 1. That reads fine for the 2-node example, but the Nodes row can now size a cluster past 2: Kimi-K3 on H100 carries a 32-GPU floor (`strategy_min_gpus: { h100: 32 }`), so it renders at **4 nodes** — and the command block still showed only `Head` + `Node 1`, leaving the reader to hand-edit `--node-rank` for nodes 2 and 3. Replaying rank 1 on every follower hangs the mp rendezvous, so the two-tab rendering was actively misleading rather than merely incomplete.

## Change

- `buildArgs(role, nodeRole, nodeRank)` takes the follower's rank; `followerRanks(n)` = `[1 … n-1]`.
- Synthesis emits **one command per node**: `workerCommands[]` / `workerArgvs[]` alongside the existing singular keys.
  - TP / TEP / TP+PP → `--node-rank <rank>` (+ `--headless` for rank > 0)
  - DEP → `--data-parallel-start-rank = rank × dp_local` (was always one node's worth, i.e. correct for rank 1 only)
- The command block renders a tab per node — `Head · Node 1 · … · Node N-1` — for both plain multi-node and Mooncake instances that span >1 node. Endpoint substitution (`$HEAD_IP` → user value) now applies to every rank, not just rank 1.
- JSON API carries the full list: `worker_commands[]` / `worker_argvs[]` (+ `worker_docker_*s[]`, and `vllm_worker_*s[]` under `kv_store_lb`).

The singular `worker_command` / `worker_argv` keep their old meaning (rank 1), so nothing existing breaks — but agents should loop the plural arrays.

## Verification

- `node scripts/build-recipes-api.mjs` → `✓ JSON API: 151 models …`; `next build` prerenders all pages clean.
- Kimi-K3 / H100 / 4 nodes: `head rank=0`, then `Node 1/2/3` at `--node-rank 1/2/3` (`--headless` on each); DEP renders `start_rank 8 / 16 / 24`; Mooncake instances render follower ranks 1/2/3.
- Diffed **all 13,813 generated JSON files** against `main`: **zero changed values**, only the new keys — the change is purely additive for every existing 2-node rendering.


esmeetu:
Pushed a second commit (d035291) fixing a related node-count leak the per-node tabs made visible: the `strategy_min_gpus` floor effect only ever **grew** the count, so 4 nodes grown for H100's 32-GPU floor stuck when switching to H200 (whose Nodes row only offers 1/2) and the builder kept rendering 4 tabs there. The effect now reconciles both directions — a count outside the current hardware's node options snaps back to `max(2, this hardware's own floor)` — and is guarded off for `pd_cluster` and non-scalable workstations. Generated JSON is unaffected (the API computes per-hardware from scratch; H200 was already 2 nodes there).

PR #716, opened 2026-07-31
Gemma-4 26B-A4B recipe update

prodney-dev:
Add recipe for DGX Station, RTX Pro 6000 and RTX 5090

### DIFF
--- a/models/Google/gemma-4-26B-A4B-it.yaml
+++ b/models/Google/gemma-4-26B-A4B-it.yaml
@@ -34,6 +34,9 @@ meta:
     ironwood: verified
     xeon6: verified
     dgx_spark_gb10: verified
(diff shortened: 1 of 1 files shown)


esmeetu:
Can we just make MRV2 as default? set it into `base_env: {}`

prodney-dev:
> Can we just make MRV2 as default? set it into `base_env: {}`

@esmeetu I can make the change. The main reason for adding it was this bug https://github.com/vllm-project/vllm/issues/48848 which we have observed on the edge devices. I haven't tested across the other devices so I was a bit hesitant to set it as default. Let me know if you still prefer it.

esmeetu:
LGTM, Let's update it once the bug fixed!

PR #736, opened 2026-08-04
Add NVIDIA Nemotron 3.5 Lightning

faradawn:
Adds a recipe for NVIDIA Nemotron 3.5 Lightning covering the published NVFP4 and BF16 checkpoints on H100, DGX Spark and DGX Station, with the full 1M context, MTP / DSpark / DFlash speculative decoding modes, and the Hopper-specific Humming MoE and linear backends alongside the FlashInfer Mamba cache flags the model requires.


models/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16.yaml
```diff
@@ -0,0 +1,291 @@
...
+    --mamba-cache-philox-rounds 5 \
+    --mamba-ssu-algorithm horizontal \
+    --reasoning-parser nemotron_v3 \
+    --enable-auto-tool-choice \
+    --tool-call-parser qwen3_xml \
+    --speculative-config '{"method":"dspark","model":"nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-NVFP4-DSpark","num_speculative_tokens":3}'
+  ```
+
+  Drop the `--speculative-config` line for the max-throughput configuration, or swap the
+  method for `mtp` / `dflash` (DFlash uses the
```
shaunkotek: this sounds like dflash is for max throughput which is not.

models/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16.yaml
```diff
@@ -0,0 +1,291 @@
...
+
+  128 GB of unified memory is shared with the host, so cap `--gpu-memory-utilization`
+  well below the default. `--load-format fastsafetensors` cuts the load time of the
+  NVFP4 weights on the Spark's slower storage path.
+
+  ```bash
+  export VLLM_ALLOW_LONG_MAX_MODEL_LEN=1
+
+  vllm serve nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-NVFP4 \
+    --trust-remote-code \
```
shaunkotek: trust remote code is not needed anymore to all nemotron models

models/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16.yaml
```diff
@@ -0,0 +1,291 @@
...
+    --moe-backend marlin \
+    --kv-cache-dtype fp8 \
+    --max-model-len 1048576 \
```
shaunkotek: this is part of the model and not needed

models/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16.yaml
```diff
@@ -0,0 +1,291 @@
...
+    --enable-mamba-cache-stochastic-rounding \
+    --mamba-cache-philox-rounds 5 \
+    --reasoning-parser nemotron_v3 \
+    --enable-auto-tool-choice \
+    --tool-call-parser qwen3_xml \
+    --load-format fastsafetensors \
+    --speculative_config.method dspark \
+    --speculative_config.model nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-NVFP4-DSpark \
+    --speculative_config.num_speculative_tokens 3 \
+    --speculative_config.max_model_len 1048576
```
shaunkotek: same

models/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16.yaml
```diff
@@ -0,0 +1,291 @@
...
+
+  vllm serve nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-NVFP4 \
+    --host 0.0.0.0 \
+    --port 8000 \
+    --tensor-parallel-size 1 \
+    --trust-remote-code \
+    --moe-backend marlin \
+    --kv-cache-dtype fp8 \
+    --attention-backend flashinfer \
+    --max-model-len 1048576 \
```
shaunkotek: same

models/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16.yaml
```diff
@@ -0,0 +1,291 @@
...
+  ### 1x DGX Station (GB300) — NVFP4 with MTP
+
+  ```bash
+  export VLLM_ALLOW_LONG_MAX_MODEL_LEN=1
+
+  vllm serve nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-NVFP4 \
+    --host 0.0.0.0 \
+    --port 8000 \
+    --tensor-parallel-size 1 \
+    --trust-remote-code \
```
shaunkotek: same

models/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16.yaml
faradawn: Good catch — reworded. DSpark is now called out as the lowest-latency option, and "maximum throughput" is explicitly tied to dropping `--speculative-config` altogether rather than to DFlash.

models/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16.yaml
faradawn: Removed `--trust-remote-code` from every command and from the recipe's `base_args`, so the builder no longer emits it either.

models/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16.yaml
faradawn: Dropped `--max-model-len` — the 1M context comes from the checkpoint. `VLLM_ALLOW_LONG_MAX_MODEL_LEN` went with it since it only existed to allow that override.

models/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16.yaml
faradawn: Dropped `--speculative_config.max_model_len` too.

models/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16.yaml
faradawn: Dropped here as well.

models/nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16.yaml
faradawn: Removed.

faradawn:
@shaunkotek all six addressed in 66101d9 — `--trust-remote-code` is gone from the guide and from the recipe's `base_args`, `--max-model-len` / `--speculative_config.max_model_len` (and the `VLLM_ALLOW_LONG_MAX_MODEL_LEN` that only existed to allow them) are dropped, and the throughput sentence no longer reads as if DFlash were the throughput option.

On the @gemini-code-assist review: those comments are against the pre-rename file and I don't think any of the three high-priority ones hold, so I've left them as-is. Checking against vLLM main:

- `qwen3_coder` and `qwen3_xml` are both registered tool parsers (`vllm/tool_parsers/__init__.py`). The recipe uses `qwen3_xml`, which is what we validated on the nightly.
- `--mamba-ssm-cache-dtype` is a real flag (`vllm/engine/arg_utils.py`, alongside the separate `--mamba-cache-dtype`); it scopes the dtype to the SSM state specifically, which is what we want here.
- `dflash` is a first-class speculative method (`vllm/config/speculative.py`) and the draft repo goes in `model`, not `draft_model`.

Happy to revisit if I've misread any of those.

eugr:
@faradawn DGX Spark looks good

faradawn:
Thanks @prodney-dev @eugr for the review.

PR #738, opened 2026-08-04
[ROCm][Kimi-K3] add VLLM_ROCM_USE_AITER_MOE_SITUV2_A8W4 for the a8w4 …

fanxingran:
…MoE path

vllm-project/vllm#50582 renamed the switch; the pinned kimi-k3 image still reads AITER_SITUV2_A8W4, so set both. Mixing them corrupts MoE output.


fanxingran:
cc @Fangzhou-Ai @hongxiayang — sorry to ping you directly, and please
correct me if I've misread anything here.
vllm-project/vllm#50582 renamed the a8w4 SiTU MoE switch to
`VLLM_ROCM_USE_AITER_MOE_SITUV2_A8W4`, while the pinned `kimi-k3` image (Jul 27)
still reads only `AITER_SITUV2_A8W4`. This PR keeps both so either build lands on
the same fp8 kernels. As far as I can tell, the old name alone on a newer build is
silently wrong: vLLM preshuffles w13 separated while AITER picks the `_gui`
interleaved kernels.
Verified on 8x MI355X (TP8): kernels resolve to
`flydsl_moe1_afp8_wfp4_bf16_*_gui_fp8` and the output looks clean. Happy to drop
the `AITER_*` pair once the image is refreshed — please let me know if you'd
prefer that now, or if there's a validation you'd like me to run first. Thanks
for taking a look!

models/moonshotai/Kimi-K3.yaml
```diff
@@ -193,6 +193,11 @@ hardware_overrides:
     extra_env:
       VLLM_ROCM_USE_AITER: "1"
       SAFETENSORS_FAST_GPU: "1"
+      # a8w4 (fp8 activation) SiTU MoE. vllm-project/vllm#50582 renamed the
+      # switch; the pinned kimi-k3 image still reads the AITER_* pair, so set
+      # all three — and flip them together, a mixed state silently corrupts
+      # MoE output.
+      VLLM_ROCM_USE_AITER_MOE_SITUV2_A8W4: "1"
       AITER_SITUV2_A8W4: "1"
```
tjtanaa: @fanxingran let's remove this `AITER_SITUV2_A8W4: "1"` since it has been deprecated.

hongxiayang:
kimi-k3 day 0 image and vllm rocm nightly image use different environment variables for the control to use the a8w4 route. The env and docker image need to align

models/moonshotai/Kimi-K3.yaml
fanxingran: @tjtanaa Agreed the new name is canonical after #50582. I kept both only
because the pinned `vllm/vllm-openai-rocm:kimi-k3` day-0 image still reads just
`AITER_SITUV2_A8W4`, and there the new name alone fails silently — wrong MoE
output, no error. Same misalignment @hongxiayang notes above.

Is the `kimi-k3` tag already rebuilt on a vLLM that reads the new name? If so
I'll drop the old one right away; if not, I'd keep both until it is refreshed. Thanks a lot!

PR #741, opened 2026-08-04
Add EximiusLabs/fusion-embedding-2-2b-preview (multimodal embedding: text, image, video, audio)

abtonmoy:
Adds a recipe for [EximiusLabs/fusion-embedding-2-2b-preview](https://huggingface.co/EximiusLabs/fusion-embedding-2-2b-preview), a 2.8B multimodal embedding model that places text, images, video and audio in one 2048-d space (frozen Qwen3-VL-Embedding-2B base + frozen Qwen2.5-Omni audio tower + token-gated adapters).

The model serves through an out-of-tree plugin registered via the `vllm.general_plugins` entry point; it ships inside the `fusion-embedding` PyPI package, so the install is `uv pip install 'vllm[audio]==0.26.0' fusion-embedding` and the launch is a plain `vllm serve EximiusLabs/fusion-embedding-2-2b-preview --runner pooling` (no trust-remote-code, no extra flags). Smoke-tested against vllm==0.26.0.

Notes for review:

- `node scripts/build-recipes-api.mjs` passes and renders `public/EximiusLabs/fusion-embedding-2-2b-preview.json`.
- Every number in the recipe is measured: parameter counts summed from safetensors metadata, context length from the base config, parity figures from executed comparison runs against the reference implementation (fp32 text cosine 0.999998+, image 0.999986+, audio 0.999999+; the served HTTP endpoint verified end to end).
- Single-GPU, `single_node_tp` with `tp: 1`, eager execution forced by the plugin (documented in the guide).

Happy to adjust tags, difficulty, or guide structure to fit conventions.

### DIFF
--- a/models/EximiusLabs/fusion-embedding-2-2b-preview.yaml
+++ b/models/EximiusLabs/fusion-embedding-2-2b-preview.yaml
@@ -1,0 +1,173 @@
+meta:
+  title: "Fusion Embedding 2 (2B preview)"
+  slug: "fusion-embedding-2-2b-preview"
+  provider: "Eximius Labs"
+  description: "Multimodal embedding model that places text, images, video and audio in one 2048-d space, built on a frozen Qwen3-VL-Embedding-2B base with a frozen Omni audio tower and token-gated adapters."
+  date_added: 2026-08-04
+  date_updated: 2026-08-04
+  difficulty: intermediate
+  tasks:
+    - embedding
+    - multimodal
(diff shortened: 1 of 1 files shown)


abtonmoy:
Addressed the review: bumped `vram_minimum_gb` to 8 (matches the guide's prerequisites line and real hardware tiers) and wrapped the offline `llm.embed` input in a list, both in 3e50af9. Kept `uv pip install` for the install commands since that is the dominant convention across the existing recipes (285 occurrences vs 11 plain `pip install`); happy to switch if maintainers prefer otherwise. Validator still passes.

PR #744, opened 2026-08-04
Add RedHatAI/Kimi-K3-NVFP4 recipe

kylesayrs:
## Summary

- Adds `models/RedHatAI/Kimi-K3-NVFP4.yaml` — a compressed-tensors NVFP4 quantization of Kimi-K3 by Red Hat AI
- Mirrors `moonshotai/Kimi-K3` deployment (same MoE strategies, Blackwell overrides, features: tool calling, reasoning, DSpark spec decoding, text-only)
- Key differences from the base recipe:
  - `precision: nvfp4`, `vram_minimum_gb: 1976` (sized from real 1645.9 GB safetensors checkpoint)
  - NVIDIA-only hardware (b200/b300/gb200/gb300 verified); no AMD section since NVFP4 is CUDA-specific
  - `nightly_required: true` — vLLM CT support for Kimi-K3 (PR vllm-project/vllm#50500) merged 2026-07-31, after v0.26.0

## Benchmarks ##

From @mgoin
Screenshot 2026-07-31 at 3 37 56 PM


models/RedHatAI/Kimi-K3-NVFP4.yaml
mgoin, line 1: this doesn’t seem right as adding it as a completely new model, it should be added as a variant on the base moonshotai/Kimi-K3 model card

mgoin (approved):

PR #751, opened 2026-08-05
[DSV4] Use probabilistic draft sampling

benchislett:
It's ~15% faster according to my testing on the latest DSV4-Flash-0731

### DIFF
--- a/models/deepseek-ai/DeepSeek-V4-Flash.yaml
+++ b/models/deepseek-ai/DeepSeek-V4-Flash.yaml
@@ -82,7 +82,7 @@ features:
               - '{"method":"mtp","num_speculative_tokens":1}'
       dspark:
         label: "DSpark"
-        description: "DSpark drafting, 7 draft tokens, greedy. Requires the DSpark or 0731 checkpoint (Variant row)."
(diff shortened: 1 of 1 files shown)


models/deepseek-ai/DeepSeek-V4-Flash.yaml
```diff
@@ -82,15 +82,15 @@ features:
               - '{"method":"mtp","num_speculative_tokens":1}'
       dspark:
         label: "DSpark"
-        description: "DSpark drafting, 7 draft tokens, greedy. Requires the DSpark or 0731 checkpoint (Variant row)."
+        description: "DSpark drafting, 7 draft tokens, probabilistic sampling. Requires the DSpark or 0731 checkpoint (Variant row)."
```
tlrmchlsmth, line 85: should we update the model_id on line 25 then?

mgoin (approved):

models/deepseek-ai/DeepSeek-V4-Flash.yaml
esmeetu: For consistency, the model ID in the YAML follows the model name. That said, I think DeepSeek's preview model (DeepSeek-V4-Flash) should probably use the `-Preview` suffix. Anyway I have made 0731 by default in `variant`, so i think it's fine now.

PR #753, opened 2026-08-06
Add new recipes for Intel XPU(B60, B70) 

Dboyqiao:
for nemotron-3-nano, gemma-4 and qwen3.5/qwen3.6


Dboyqiao:
@jikunshang Could you help to review?

models/Google/gemma-4-12B-it.yaml
```diff
@@ -74,7 +76,27 @@ compatible_strategies:
   - single_node_tp
   - multi_node_tp
 
-hardware_overrides: {}
+hardware_overrides:
+  xpu:
+    extra_env:
+      VLLM_ALLOW_LONG_MAX_MODEL_LEN: "1"
+      VLLM_WORKER_MULTIPROC_METHOD: "spawn"
```
jikunshang: i feel we don't need these params any more

models/Google/gemma-4-12B-it.yaml
```diff
@@ -74,7 +76,27 @@ compatible_strategies:
...
+    extra_args:
+      - "--enforce-eager"
+      - "--attention-backend"
```
jikunshang, line 83: it's default option, we can remove.

models/Google/gemma-4-12B-it.yaml
```diff
@@ -74,7 +76,27 @@ compatible_strategies:
...
+      - "--trust-remote-code"
+      - "--gpu-memory-utilization"
+      - "0.92"
+      - "--no-enable-prefix-caching"
+      - "--max-num-batched-tokens"
+      - "8192"
+      - "--max-model-len"
+      - "5120"
+      - "--block-size"
+      - "64"
```
jikunshang: can remove

models/Google/gemma-4-12B-it.yaml
yintong-lu: I kept `--attention-backend flash_attn` intentionally. For Gemma 4 on XPU, if the attention backend is not explicitly set, vLLM treats Gemma 4 as a multimodal prefix-LM model and the XPU backend selector falls back to Triton attention because Flash Attention does not support the multimodal prefix-LM mask path. 
Removed other args as per suggested.

jikunshang: got it thanks.

jikunshang:
please fix dco issue.

jikunshang:
please fix dco issue and remove source oneapi part.

Dboyqiao:
> please fix dco issue and remove source oneapi part.

Done

models/Google/gemma-4-12B-it.yaml
rogerxfeng8: > I kept `--attention-backend flash_attn` intentionally. For Gemma 4 on XPU, if the attention backend is not explicitly set, vLLM treats Gemma 4 as a multimodal prefix-LM model and the XPU backend selector falls back to Triton attention because Flash Attention does not support the multimodal prefix-LM mask path. Removed other args as per suggested.

Is it similar behavior for other vendors (need to explicit specify the attn backend)? seems not user friendly.

models/Google/gemma-4-12B-it.yaml
```diff
@@ -74,7 +76,22 @@ compatible_strategies:
...
-hardware_overrides: {}
+hardware_overrides:
+  xpu:
+    extra_args:
+      - "--enforce-eager"
+      - "--attention-backend"
+      - "flash_attn"
+      - "--trust-remote-code"
+      - "--gpu-memory-utilization"
+      - "0.92"
```
rogerxfeng8: Is 0.92 having chances to trigger the OOM in users environment? How about to take 0.9 for tolerance?

models/Google/gemma-4-12B-it.yaml
yintong-lu: This is not a general requirement for all vendors.
The special case here is Gemma 4 on XPU. Gemma 4 is treated as a multimodal prefix-LM model, and the XPU backend selector currently falls back to Triton when `use_mm_prefix=True`, because FlasH-Attn on XPU does not support the multimodal prefix-LM mask path. 

The longer-term fix should probably be in vLLM's backend selection logic, so that text-only usage can still auto-select Flash-Attn while preserving Triton fallback for true multimodal prefix-LM inputs. 
For this recipe, I kept the explicit `--attention-backend flash_attn` to make the validated XPU text-only command deterministic and avoid unexpectedly running with Triton.

models/Google/gemma-4-12B-it.yaml
yintong-lu: yes 0.9 works.
Adjustment as been made accordingly.

jikunshang (commented):
please fix conflict

Dboyqiao:
> please fix conflict

Done

Dboyqiao:
@jikunshang conflict is fixed, please help to review

jikunshang (approved):

PR #767, opened 2026-08-09
[Model] Add Dots3Note serving recipe

KurodaKanbei:
## Summary

Add a serving recipe for Dots3Note, a 288B-parameter multimodal MoE model with approximately 17B active parameters, 512K context support, hybrid DSA/full and sliding-window attention, and a three-layer MTP head.

The recipe covers both published checkpoints:

- BF16: `dots-studio/dots-3-note-prev`
- FP8: `dots-studio/dots-3-note-prev-fp8`

## Deployment configurations

The guide documents:

- Tensor parallelism (TP)
- Tensor parallelism with expert parallelism (TP + EP)
- Data parallelism with expert parallelism (DP + EP)
- Language-model-only serving with `--language-model-only`
- Three-token MTP speculative decoding

The default commands use the FP8 checkpoint on eight NVIDIA GPUs. The BF16 checkpoint is exposed as a selectable recipe variant with its larger memory requirement documented.

The recipe uses a 256K validated serving limit and an 8K chunked-prefill budget while retaining the model’s documented 384K maximum context capability.

## Additional changes

- Register the `dots-studio` provider
- Use the Dots Studio Hugging Face organization logo
- Add links to both BF16 and FP8 model repositories
- Link the corresponding vLLM model-support PR

## Validation

Ran the complete recipe API build successfully:

- FP8 promoted model: `dots-studio/dots-3-note-prev-fp8`, FP8, 359 GB
- BF16 promoted model: `dots-studio/dots-3-note-prev`, BF16, 692 GB
- Recipe JSON generation completed without errors

### DIFF
--- a/scripts/fetch-provider-logos.mjs
+++ b/scripts/fetch-provider-logos.mjs
@@ -52,7 +52,7 @@ for (const [org, meta] of Object.entries(targets)) {
     continue;
   }
   try {
-    const avatarUrl = await fetchOrgAvatarUrl(org);
+    const avatarUrl = await fetchOrgAvatarUrl(meta.avatar_org || org);
     if (!avatarUrl) {
       console.warn(`⚠ no avatar found for ${org}`);
       failed++;
--- a/scripts/fetch-provider-logos.mjs
+++ b/scripts/fetch-provider-logos.mjs
@@ -52,7 +52,7 @@ for (const [org, meta] of Object.entries(targets)) {
     continue;
   }
   try {
-    const avatarUrl = await fetchOrgAvatarUrl(meta.avatar_org || org);
+    const avatarUrl = await fetchOrgAvatarUrl(org);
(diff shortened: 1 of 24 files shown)


scripts/fetch-provider-logos.mjs
```diff
@@ -52,7 +52,7 @@ for (const [org, meta] of Object.entries(targets)) {
     continue;
   }
   try {
-    const avatarUrl = await fetchOrgAvatarUrl(org);
+    const avatarUrl = await fetchOrgAvatarUrl(meta.avatar_org || org);
```
xianbaoqian:
```suggestion
    const avatarUrl = await fetchOrgAvatarUrl(org);
```

I don't think this change is needed

KurodaKanbei: Already resolved

PR #784, opened 2026-08-11
DGX Spark DeepSeek V4 flash recipe

prodney-dev:
Tested the recipe rendering logic and the custom note

Test screenshot


faradawn:
LGTM. @esmeetu what do you think

PR #786, opened 2026-08-11
Change DSv4 backend on DGX Spark to b12x

prodney-dev:
- Change the backends to b12x and the docker image to eugr/spark-vllm-b12x:latest
- Also update the speculative decoding config for DGX Spark
    - **command-synthesis.js** changes needed to honor the device specific override for the speculative decoding config


esmeetu:
python (pip) serve command will still be affected by this as well, is that right?

prodney-dev:
> python (pip) serve command will still be affected by this as well, is that right?

Yes, both pip and docker commands will be updated.

PR #799, opened 2026-08-13
fix: isolate tooltip provider state

eladmotola:
When moving quickly between buttons, Radix reused the shared tooltip provider’s “instant open” state. The next tooltip could render before Popper recalculated its position, briefly appearing in the previous tooltip’s location.
The fix gives each InfoTip its own provider, preventing state from being carried between buttons.

### DIFF
--- a/src/components/ui/tooltip.jsx
+++ b/src/components/ui/tooltip.jsx
@@ -41,12 +41,14 @@ const TooltipContent = React.forwardRef(function TooltipContent(
 function InfoTip({ content, children, side, align, sideOffset, asChild = true }) {
   if (!content) return children;
   return (
-    <Tooltip>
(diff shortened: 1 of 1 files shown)


eladmotola:
I agree with Gemini that isolating providers regresses the warm-up behavior. However, key={content} on each Tooltip will not remount a tooltip when moving between different triggers, because every InfoTip already owns a distinct Tooltip instance.
I also tried to test disableHoverableContent but we are unable the place the mouse over the tooltip.... So I'm not sure what direction we should take. All I know that right now the tooltip experience is not the best

eladmotola:
Hey @esmeetu 
Sorry for the tagging. I hope you could review my change

esmeetu:
LGTM! Thanks.

PR #805, opened 2026-08-14
[Recipe] Add Qwen3.8-27B

xianbaoqian:
Dense 27B sibling of Qwen3.8-2.4T-A95B on the same hybrid-attention backbone: linear attention on 48 of 64 layers, a vision tower, a built-in MTP draft head, 262K native context extensible to 1M. Structure and wording follow the 2.4T recipe.

Verified on GB300 (vLLM 0.1.dev19754+g3a0914114): BF16 and FP8 at TP4, NVFP4 at TP1, 1M context with correct needle retrieval at 193K / 385K / 867K, working tool calls, MTP-3 acceptance up to 92.2%. No throughput sweep was run, so the headline is a capacity statement rather than a measurement.

Differences from the 2.4T that are load-bearing:

- TP is 1/2/4/8 only. 24 attention heads exclude 16 and 32; the linear-attention key projection (2048 = 2^11) excludes 3, 6, 12 and 24. Copying the 2.4T's "1/2/4/8/16/32" would ship commands that cannot start.
- single_node_tp only. Multi-node TP would set TP = 16 on the smallest 2-node layout, which 24 heads do not divide, and nothing here needs a second node: 51.7 GiB of BF16 weights fit one GPU.
- The 1M context override nests under text_config, where the text-only 2.4T takes it flat. The flat form silently does nothing on this checkpoint.
- Dense, so no expert-parallel or PD strategies, and no MXFP4 variant: vLLM implements MXFP4 for MoE experts only and returns UnquantizedLinearMethod for every LinearBase.
- min_vllm_version is numeric, not "nightly". maxVersion() parses each field through parseInt(...) || 0, so "nightly" scores 0.0.0 and loses to the kv_store YAMLs' 0.21.0, understating the requirement in the Install header whenever a Mooncake KV-offload option is selected.
- The docker pin is NVIDIA-only. The same-named rocm tag is an AMD build of the 2.4T text model and has never been exercised here.

The vision tower loads and the placeholders are wired, but no image or video request has been run against this recipe; the guide says so rather than implying multimodal coverage.

### DIFF
--- a/models/Qwen/Qwen3.8-27B.yaml
+++ b/models/Qwen/Qwen3.8-27B.yaml
@@ -1,0 +1,238 @@
+meta:
+  title: "Qwen3.8-27B"
+  slug: "qwen3.8-27b"
+  provider: "Qwen"
+  description: "27B-parameter dense hybrid-attention model with linear attention on 48 of 64 layers, a vision tower, a built-in MTP draft head, 262K native context window and extensible to 1M context"
+  date_added: 2026-08-14
+  date_updated: 2026-08-14
+  difficulty: intermediate
+  tasks:
+    - multimodal
+    - text
+  performance_headline: "Fits one Blackwell GPU in every precision: NVFP4 in 24.6 GiB, 6.6M KV tokens at 1M context"
+  related_recipes:
+    - "Qwen/Qwen3.8-2.4T-A95B"
+  # `verified` = run end-to-end on that GPU. gb300 earned it on 2026-08-14: BF16 and FP8 at
+  # TP4, NVFP4 at TP1 on vLLM 0.1.dev19754+g3a0914114, correct generations at 1M context,
+  # and working tool calls. Everything else stays absent — the honest silent default.
+  hardware:
+    gb300: verified
+
+model:
+  model_id: "Qwen/Qwen3.8-27B"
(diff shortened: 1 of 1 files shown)


VoyagerXvoyagerx:
Cheers! 🚀

models/Qwen/Qwen3.8-27B.yaml
```diff
@@ -0,0 +1,238 @@
...
+  # requirement. 0.27.2 is the in-flight nightly (PyPI stops at 0.27.1), so
+  # `nightly_required` still drives the wheel index — the version just renders honestly now.
+  #
+  # Base serving is code-identical on stable 0.27.1, but was never run there. The floor
+  # exists for `spec_decoding`: the GDN speculative-gate fix (vllm-project/vllm#51812) is in
+  # no released tag. Note the tested image predates that PR and gets the same correctness
+  # from #51674 instead — neither is a superset of the other, so treat 0.27.2 as a
+  # correctness floor, not a reproduce-these-numbers pin.
+  min_vllm_version: "0.27.2"
+  nightly_required: true
```
Isotr0py, line 36: Actually, we don't need nightly version for this model because it's same as Qwen3.5.

PR #808, opened 2026-08-14
updated kimi-k3 dspark recipe

shanjiaz:
Updated the recipe to use updated kimi-k3 dspark model https://huggingface.co/RedHatAI/Kimi-K3-speculator.dspark. 

This model uses sliding window attention and performs well for long context prompts up to 1m.


models/moonshotai/Kimi-K3.yaml
mgoin, line 32: We have to update this to include the current latest nightly

mgoin (approved):

PR #814, opened 2026-08-15
Kimi-K3: document alternative MLA prefill backends and DCP decode layout

Ankur-singh:
## Summary

Two fixes to the Kimi-K3 recipe, from reconciling it against a real B300 deployment.

1. **Expose decode context parallelism as an opt-in command-builder feature.**
   `--decode-context-parallel-size`, `--dcp-comm-backend a2a` and
   `--attention-backend TOKENSPEED_MLA` were undocumented. They are now a `dcp`
   feature listed in `opt_in_features`, scoped to `single_node_tp` because the
   pinned DCP size of 8 only divides a tensor-parallel size of 8 or more. The
   feature also carries `mla_prefill_backend: TRTLLM_RAGGED`, which — because
   features emit last and dedupe is last-wins — cleanly replaces the Blackwell
   baseline's `flashinfer` value rather than duplicating the flag.

2. **The FP8 KV note no longer reads as if `flashinfer` were mandatory.** vLLM
   0.27.1 registers three MLA prefill backends in
   `vllm/v1/attention/backends/mla/prefill/registry.py` — `FLASHINFER`,
   `TRTLLM_RAGGED` and `TOKENSPEED_MLA`. `flashinfer` stays the recommended
   default and is still what the Blackwell profile emits with the feature off.

## Validation

`node scripts/build-recipes-api.mjs` passes: 160 models, 153 promoted variants, 9
strategies. The four variant-collision warnings are pre-existing and unrelated.

Rendering `single_node_tp` on `b300` through `resolveCommand`:

| | `--attention-config` | `--attention-backend` | DCP |
|---|---|---|---|
| feature off | `…"mla_prefill_backend":"flashinfer"` | *(none)* | *(none)* |
| feature on | `…"mla_prefill_backend":"TRTLLM_RAGGED"` | `TOKENSPEED_MLA` | `--decode-context-parallel-size 8` |

Exactly one `--attention-config` in both cases — no stale duplicate — and the
default profile is unchanged when the feature is off.

## Context

Raised while reviewing
[SemiAnalysisAI/InferenceX#2616](https://github.com/SemiAnalysisAI/InferenceX/pull/2616),
which serves Kimi-K3 on B300 with TP8 x DCP8, `TOKENSPEED_MLA` and
`mla_prefill_backend: TRTLLM_RAGGED`. That submission could not be reconciled
against this recipe: two of those flags were undocumented here and the third read
as contradicting the FP8 KV guidance. InferenceX requires every single-node
vLLM submission to be documented in an upstream recipe, so this PR closes that
gap for the DCP deployment shape.

cc @faradawn for review


faradawn (commented):
Thanks for adding DCP and providing the commands. 

First, Can we modify the command builder, instead of the markdown section? E.g. 
```
features:
  # ... existing features ...
  dcp:
    args:
      - "--decode-context-parallel-size"
      - "8"
      - "--dcp-comm-backend"
      - "a2a"
      - "--attention-backend"
      - "TOKENSPEED_MLA"
      - "--attention-config"
      - '{"use_prefill_query_quantization":true,"mla_prefill_backend":"TRTLLM_RAGGED"}'

opt_in_features:
  - spec_decoding
  - text_only
  - dcp          # <-- add
```

Second, does K3's DCP path requires vLLM >= 0.27.2 (instead of 0.27.1);

Third, do we need to set `VLLM_USE_DIRECT_DCP_A2A=1`, `VLLM_USE_DIRECT_DCP_Q_GATHER=1` and `VLLM_USE_DIRECT_DCP_KV_GATHER=1` 

Please let me know if the comments make sense!

Ankur-singh:
Thanks for the detailed review — the command-builder suggestion is much better than the
markdown section I had, and the two questions sent me digging in useful directions. All
three are addressed in `7d467ef`.

**1. Command builder instead of markdown** — done, in the shape you sketched: a `dcp`
entry under `features:` plus `dcp` in `opt_in_features`, replacing the markdown bullet.
Two small adjustments I'd like your read on:

- I scoped it with `strategies: [single_node_tp]`. Since `--decode-context-parallel-size`
  is pinned to `8` in the args, it only divides a tensor-parallel size of 8 or more, so
  offering the toggle on the TP4 tray strategies would emit a command that can't run.
  Happy to widen it if you'd prefer it available more broadly.
- I wanted to be sure the `--attention-config` in your snippet composed correctly, since
  the `blackwell` hardware override already emits that flag with
  `mla_prefill_backend: flashinfer`. `command-synthesis.js` notes that features emit last
  and `dedupeArgs` is last-wins, and rendering `single_node_tp` on `b300` through
  `resolveCommand` confirms it behaves as you intended:

  | | `--attention-config` | `--attention-backend` | DCP |
  |---|---|---|---|
  | feature off | `…"mla_prefill_backend":"flashinfer"` | *(none)* | *(none)* |
  | feature on | `…"mla_prefill_backend":"TRTLLM_RAGGED"` | `TOKENSPEED_MLA` | `--decode-context-parallel-size 8` |

  Exactly one `--attention-config` in both cases, and the default profile is untouched
  when the feature is off.

**2. Does the DCP path need >= 0.27.2?** — I went looking for 0.27.2 and couldn't find a
tag for it; `v0.27.1` still appears to be the latest release, so I've left
`min_vllm_version` alone for now. Checking what's actually in that tag rather than main:

- `--decode-context-parallel-size` and `--dcp-comm-backend` — both in
  `vllm/engine/arg_utils.py` at `v0.27.1`
- `TOKENSPEED_MLA` and `TRTLLM_RAGGED` — both registered in
  `vllm/v1/attention/backends/mla/prefill/registry.py` at `v0.27.1`
- `vllm/v1/attention/ops/dcp_alltoall.py` — present at `v0.27.1`

So the flags themselves do appear to be in the release. What *isn't* in 0.27.1 are the
`VLLM_USE_DIRECT_DCP_*` knobs (more on those below), which may well be what prompted your
question. That said, you have far more context on the K3 integration than I do — if
there's a K3-specific DCP fix that landed after 0.27.1 and matters for correctness rather
than just performance, I'd much rather gate the feature on it explicitly than leave the
recipe implying 0.27.1 is enough. If you can point me at the commit I'll add a note or a
per-feature version gate.

**3. `VLLM_USE_DIRECT_DCP_A2A` / `_Q_GATHER` / `_KV_GATHER`** — I've left these out of the
feature args for now, and wanted to explain the reasoning in case you'd rather they were
included. In `vllm/envs.py` all three are `bool | None = None`, with the comment above
them reading *"Direct DCP ops default on when applicable; set to 1 to enforce or 0 to
disable."* Being tri-state and auto-enabling where applicable, setting `=1` in the feature
would pin the direct path for every user rather than switch on something otherwise off —
and they aren't present in `v0.27.1` at all. I've mentioned them in the guide note so
anyone pinning the path for reproducible benchmarking knows they exist, but I'm glad to
move them into the args if you'd prefer the explicit form.

Happy to iterate on any of this.

models/moonshotai/Kimi-K3.yaml
```diff
@@ -81,10 +81,26 @@ features:
     description: "Skip the vision encoder for text-only workloads. Mutually exclusive with encoder_parallel."
     args:
       - "--language-model-only"
+  dcp:
+    description: "Decode context parallelism: shard the decode KV cache across the tensor-parallel ranks for decode-heavy long-context serving. The DCP size must divide the tensor-parallel size, so the args below assume the TP8 single-node layout. Pairs the TOKENSPEED_MLA decode backend with TRTLLM_RAGGED MLA prefill under FP8 KV."
```
ivanium:
```suggestion
  decode_context_parallelism:
    description: "Decode context parallelism (DCP): shard the decode KV cache across the tensor-parallel ranks for decode-heavy long-context serving. The DCP size must divide the tensor-parallel size, so the args below assume the TP8 single-node layout. Pairs the TOKENSPEED_MLA decode backend with TRTLLM_RAGGED MLA prefill under FP8 KV."
```

models/moonshotai/Kimi-K3.yaml
```diff
@@ -81,10 +81,26 @@ features:
...
+      - "a2a"
+      - "--attention-backend"
+      - "TOKENSPEED_MLA"
+      - "--attention-config"
+      - '{"use_prefill_query_quantization":true,"mla_prefill_backend":"TRTLLM_RAGGED"}'
 
 opt_in_features:
   - spec_decoding
   - text_only
+  - dcp
```
ivanium:
```suggestion
  - decode_context_parallelism
```

models/moonshotai/Kimi-K3.yaml
```diff
@@ -360,4 +376,5 @@ guide: |
...
+  - **Decode context parallelism**: enable the `dcp` opt-in feature in the command builder to shard the decode KV cache across the tensor-parallel ranks for decode-heavy long-context serving. It emits `--decode-context-parallel-size 8 --dcp-comm-backend a2a --attention-backend TOKENSPEED_MLA` with `mla_prefill_backend: TRTLLM_RAGGED`, and is scoped to the TP8 single-node profile because the DCP size must divide the tensor-parallel size. On builds that carry them, the `VLLM_USE_DIRECT_DCP_A2A`, `VLLM_USE_DIRECT_DCP_Q_GATHER` and `VLLM_USE_DIRECT_DCP_KV_GATHER` knobs default to auto and turn the direct DCP ops on where applicable; set them to `1` only to pin that path explicitly, or `0` to disable it.
```
ivanium:
```suggestion
  - **Decode context parallelism (DCP)**: enable the `dcp` opt-in feature in the command builder to shard the decode KV cache across the tensor-parallel ranks for decode-heavy long-context serving. It emits `--decode-context-parallel-size 8 --dcp-comm-backend a2a --attention-backend TOKENSPEED_MLA` with `mla_prefill_backend: TRTLLM_RAGGED`, and is scoped to the TP8 single-node profile because the DCP size must divide the tensor-parallel size. On builds that carry them, the `VLLM_USE_DIRECT_DCP_A2A`, `VLLM_USE_DIRECT_DCP_Q_GATHER` and `VLLM_USE_DIRECT_DCP_KV_GATHER` knobs default to auto and turn the direct DCP ops on where applicable; set them to `1` only to pin that path explicitly, or `0` to disable it.
```

ivanium (approved):
LGTM. Left a cosmetic suggested change.

faradawn:
Thank you @ivanium for the comments - addressed

faradawn (approved):

PR #816, opened 2026-08-17
add Qwen3 1.7,8,30B-A3B model support for Xeon

louie-tsai:
## Summary

Add Intel Xeon 6 recipes for:

- `Qwen/Qwen3-1.7B`
- `Qwen/Qwen3-8B`
- `Qwen/Qwen3-30B-A3B`

The configurations are based on validated Xeon 6 (GNR) BF16 serving runs.

This PR adds:

- `xeon6: verified`
- CPU install and Docker guidance
- CPU tuning:
  - `--max-num-batched-tokens 16384`
  - `--gpu-memory-utilization 0.8`
- Qwen3 tool-calling and reasoning feature definitions consistent with the existing Qwen3 recipes

## Validation

All three validation rows passed using vLLM `releases/v0.21.0` with 1024-token input / 1024-token output workloads:

- `Qwen/Qwen3-1.7B`: PASS
- `Qwen/Qwen3-8B`: PASS
- `Qwen/Qwen3-30B-A3B`: PASS

## Differences from the validation Excel

Some benchmark settings are intentionally not carried into the portable recipes:

- `dtype=bfloat16`: represented by the default `precision: bf16` variant; no explicit `--dtype` argument is required.
- `--trust-remote-code`: removed because Qwen3 and Qwen3MoE have native vLLM support.
- `--max-num-seqs=256`: removed because it is benchmark tuning; the deployment recipes leave sequence concurrency to vLLM defaults or user configuration.
- `--no-enable-prefix-caching`: removed because it is a benchmark-control setting and is not required for general serving; the random validation workload does not use a shared prefix.
- TP values (`TP=1` / `TP=2`): not fixed; users choose TP based on CPU topology and deployment requirements.
- `DP=6` / `DP=2` and router DP: out of scope for these model recipes.
- `CPU_VISIBLE_MEMORY_NODES` and explicit CPU binding: removed because they are host/topology specific.
- `VLLM_ENGINE_ITERATION_TIMEOUT_S=600`: removed because it is a benchmark/repro stability setting rather than a general serving requirement.
- Legacy launcher settings such as `VLLM_RPC_TIMEOUT` and `VLLM_ALLOW_LONG_MAX_MODEL_LEN` are not needed.

### DIFF
--- a/models/Qwen/Qwen3-1.7B.yaml
+++ b/models/Qwen/Qwen3-1.7B.yaml
@@ -1,0 +1,129 @@
+meta:
+  title: "Qwen3-1.7B"
+  slug: "qwen3-1.7b"
+  provider: "Qwen"
+  description: "Qwen3 1.7B dense model with hybrid thinking/non-thinking modes, validated for BF16 serving on Intel Xeon 6."
+  date_added: 2026-08-14
+  date_updated: 2026-08-14
+  difficulty: beginner
+  tasks:
+    - text
+  performance_headline: "Validated on Intel Xeon 6 (GNR) with BF16 serving"
+  related_recipes:
+    - "Qwen/Qwen3-4B"
+  hardware:
+    xeon6: verified
+
+model:
+  model_id: "Qwen/Qwen3-1.7B"
+  min_vllm_version: "0.8.5"
+  architecture: dense
+  parameter_count: "1.7B"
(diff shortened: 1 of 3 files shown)


louie-tsai:
@bigPYJ1151  please help us to review the PR

PR #820, opened 2026-08-18
[Xeon] Add remining Qwen model recipes

louie-tsai:
## Summary

Add and update Intel Xeon 6 CPU recipes for validated Qwen models, following the CPU recipe pattern used by the existing Xeon recipes and PR #789.

This PR adds new recipes for:
- `Qwen/Qwen3-14B`
- `Qwen/Qwen3-VL-30B-A3B-Instruct`
- `Qwen/QwQ-32B`

It also updates the existing `Qwen/Qwen2.5-VL-7B-Instruct` recipe with Xeon 6 support.

The CPU release validation table marks these models supported on Xeon 6. The serving repro workbook also contains model-specific serving configurations. Seven models have PASS serving results; `QwQ-32B` is CPU-supported but its 1024/1024 serving tuning row is `SLA_NOT_MET`, so no benchmark tuning values are promoted as QwQ CPU defaults.

For the PASS serving rows, the portable CPU settings retained are:

- `--max-num-batched-tokens 16384`
- `--gpu-memory-utilization 0.8`

`Qwen/Qwen1.5-0.5B-Chat` includes the CPU-validated official `Qwen/Qwen1.5-0.5B-Chat-GPTQ-Int4` checkpoint as a GPTQ INT4 variant.

## Differences from the validation Excel

Some benchmark/repro settings are intentionally not carried into the portable recipes:

- `--dtype bfloat16`: omitted; the BF16 checkpoints use vLLM's automatic dtype selection.
- `--trust-remote-code`: omitted because these Qwen model architectures have native vLLM support.
- `--max-num-seqs=256`: omitted because it is benchmark tuning and should not be prescribed independently of the user's TP/topology.
- `--no-enable-prefix-caching`: omitted because it is benchmark methodology rather than a general serving requirement.
- Fixed TP/DP and router-DP values: not prescribed; users choose parallelism based on CPU topology and deployment requirements.
- `CPU_VISIBLE_MEMORY_NODES` and explicit CPU thread binding: omitted because they are host/topology specific.
- Legacy/repro launcher settings such as `VLLM_USE_V1`, `VLLM_RPC_TIMEOUT`, `VLLM_ALLOW_LONG_MAX_MODEL_LEN`, and `VLLM_ENGINE_ITERATION_TIMEOUT_S=600`: omitted because they are not general model-serving requirements.

## Model feature/version choices

- Qwen3 dense/MoE recipes use `min_vllm_version: 0.8.5`, matching the existing Qwen3 recipe family.
[...]

PR #827, opened 2026-08-20
[Xeon] recipes for all remining supported models

louie-tsai:
## Summary

Add Intel Xeon 6 CPU recipes for the remaining CPU-supported model families that have serving BKM coverage in the validation/reproduction data, following the same normalization logic used in #820.

### New recipes

- `ibm-granite/granite-3.2-2b-instruct`
- `zai-org/glm-4-9b-hf`
- `microsoft/Phi-4-reasoning`
  - includes `TheBloke/Mistral-7B-Instruct-v0.2-AWQ` as an AWQ INT4 variant
- `microsoft/Phi-4-multimodal-instruct`
- `openai/whisper-large-v3`

## Validation

The CPU release validation table marks these model/checkpoint families supported on Intel Xeon 6. The internal serving tuning manifest and generated serving repro workbook provide model-specific BKM/reproduction rows based on vLLM CPU `releases/v0.21.0`.

Six model families have PASS serving BKM rows:

- `ibm-granite/granite-3.2-2b-instruct`
- `zai-org/glm-4-9b-hf`
- `microsoft/Phi-4-reasoning`
- `mistralai/Mistral-7B-Instruct-v0.2` (BF16 and AWQ)
- `microsoft/Phi-4-multimodal-instruct`
- `openai/whisper-large-v3`

For the PASS serving rows, the portable CPU settings retained are:

- `--max-num-batched-tokens 16384`
- `--gpu-memory-utilization 0.8`

`microsoft/Phi-4-multimodal-instruct` also retains `--trust-remote-code`, which is required for its multimodal LoRA components.

## Differences from validation Excel

Some validation/repo settings are intentionally not carried into the portable recipes:

- `--dtype bfloat16`: represented by the default BF16 variant; no explicit `--dtype` argument is needed.
- `--trust-remote-code`: omitted for models with native vLLM architecture support; retained only for Phi-4 multimodal where it is required.
- `--max-num-seqs=256`: omitted because it is benchmark/concurrency tuning rather than a model requirement.
- `--no-enable-prefix-caching`: omitted because it is benchmark methodology rather than a general serving requirement.
- Fixed TP/DP and router-DP values: not prescribed; users choose parallelism based on CPU topology and deployment requirements.
- `CPU_VISIBLE_MEMORY_NODES` and explicit CPU thread binding: omitted because they are host/topology specific.
- Repro-only timeout/launcher settings are not promoted as portable model configuration.
- Whisper's validated `native_dp` / `-dp=6` setup is documented in the guide but not hard-coded because DP is deployment topology.

## Model feature/version choices

- `min_vllm_version` is set per recipe from the earliest vLLM release
  that supports the model architecture/model path used by the recipe,
  rather than from the Xeon BKM release.
- Xeon 6 serving BKMs are tracked separately: the serving reproduction
  rows used for this PR were validated with vLLM CPU `releases/v0.21.0`.
- Mistral AWQ is modeled as a variant of the canonical Mistral-7B-Instruct-v0.2 recipe.
- Phi-4 multimodal keeps `--trust-remote-code`.
- Whisper documents the native-DP validation path without prescribing a fixed DP value.


louie-tsai:
@bigPYJ1151  could you help to review the PR? thanks!

models/deepseek-ai/DeepSeek-R1-Distill-Llama-70B.yaml
```diff
@@ -0,0 +1,81 @@
+meta:
+  title: "DeepSeek-R1-Distill-Llama-70B"
```
bigPYJ1151: This model can be categorized as a variant of meta-llama/Llama-3.3-70B-Instruct

models/ibm-granite/granite-3.2-2b-instruct.yaml
```diff
@@ -0,0 +1,80 @@
+meta:
+  title: "Granite-3.2-2B-Instruct"
+  slug: "granite-3.2-2b-instruct"
+  provider: "IBM Granite"
```
bigPYJ1151, line 4: Provider should be `IBM`? And the folder name should also be `ibm`. As this is the first model of IBM, maybe we should add a icon for it like other providers.

models/ibm-granite/granite-3.2-2b-instruct.yaml
```diff
@@ -0,0 +1,80 @@
...
+  - single_node_tp
+  - multi_node_tp
+
+hardware_overrides:
+  cpu:
+    extra_args:
+      - "--max-num-batched-tokens"
+      - "16384"
+      - "--gpu-memory-utilization"
+      - "0.8"
```
bigPYJ1151: Same as the comment in the llama PR

models/ibm-granite/granite-3.2-2b-instruct.yaml
louie-tsai: good catch, but I saw "ibm-granite" used in Huggingface too.
https://huggingface.co/ibm-granite
also github https://github.com/IBM-Granite
therefore, plan to keep ibm-granite as folder name to align with naming from Huggingface and github.

models/ibm-granite/granite-3.2-2b-instruct.yaml
louie-tsai: removed it accordingly.

models/deepseek-ai/DeepSeek-R1-Distill-Llama-70B.yaml
louie-tsai: It is derived from meta-llama/Llama-3.3-70B-Instruct, but I think it should remain a separate recipe rather than a variant. Unlike quantized variants of the same model, DeepSeek-R1-Distill-Llama-70B is a separately distilled reasoning model with its own HF repo, config/tokenizer changes, and DeepSeek-R1 reasoning behavior. 
I added meta-llama/Llama-3.3-70B-Instruct under related_recipes to capture the model lineage.

bigPYJ1151: Yes, this is a bit tricky. From perspective of model architecture, it is `LlamaForCausalLM`, same as the original meta-llama model. I would prefer to categorize it as a llama variant or not add it, to avoid potential dispute.

louie-tsai: understood. remove the recipes to avoid confusion.

bigPYJ1151 (approved):

PR #841, opened 2026-08-24
[Xeon ]fixes for xeon only variants

louie-tsai:
xeon only data types should not be listed for other architectures.


louie-tsai:
@bigPYJ1151  possible to help on reviewing the change?

bigPYJ1151 (approved):

PR #852, opened 2026-08-26
add reasoning effort for glm53flash

JaredforReal:
[reference](https://huggingface.co/zai-org/GLM-5.3-Flash/blob/main/chat_template.jinja#:~:text=%7B%25%2D%20set%20effective_reasoning_effort,%7B%25%2D%20endif%20%2D%25%7D)


esmeetu:
Thanks!

PR #857, opened 2026-08-27
[ROCm] Add GLM-5.3-Flash MI300X and MI355X validation

andyluo7:
## Summary

This is a follow-up to #853. That PR added the initial gfx950 recipe material;
this change preserves it and adds artifact-backed coverage for the pending
GLM-5.3 ROCm integration on both gfx942 and gfx950.

- adds a verified MI300X badge based on an actual 8x MI300X run;
- keeps the MI355X badge and backs it with an actual 8x MI355X run;
- intentionally does not mark MI325X verified: it shares the gfx942 code path
  but was not independently measured;
- changes the generic AMD image to ROCm nightly because the published
  `glm53-flash` image is gfx950-only;
- emits the validated TP8 / 8K / eager / AITER / Triton-MoE baseline for
  MI300X, MI325X, and MI355X; and
- documents exact accuracy results, checkpoint/source/image provenance, the
  reasoning-template caveat, and the unvalidated feature boundaries.

The required vLLM code is in the dependent ROCm PR:
https://github.com/ZJY0516/vllm/pull/5

## Hardware validation

Both runs used GSM8K 5-shot, all 1,319 samples, greedy generation, TP8,
maximum model length 8,192, eager mode, AITER, and Triton MoE.

| Hardware | Architecture | strict-match | flexible-extract |
| --- | --- | ---: | ---: |
| 8x MI300X VF | gfx942 | 0.9704321455648218 | 0.9696739954510993 |
| 8x MI355X | gfx950 | 0.974981046247157 | 0.9742228961334344 |

For both runs, all document IDs 0-1318 were present exactly once per filter,
responses were non-empty, evaluator exit status was 0, post-run health was
good, and the measured interval had no retries, JIT warnings, HIP/OOM faults,
engine deaths, or segmentation faults.

Although the evaluator supplied `enable_thinking=false`, this checkpoint's
chat template ignores that boolean and defaults to maximum reasoning effort.
These results are not presented as non-thinking scores.

This validates text generation only. It does not claim MTP, multimodal,
1M-context, graph-capture, prefill/decode, or throughput validation.

## Validation

```bash
node scripts/build-recipes-api.mjs
```

Result:

```text
✓ JSON API: 178 models (153 with recommended_command, 756 default-hw alternatives, 1439 per-hw renderings), 169 promoted variants, 9 strategies, 2 kv-store deployments, 2 platforms (7 variant collisions skipped)
```

The generated MI300X, MI325X, and MI355X JSON entries were inspected and all
emit TP8, `--max-model-len 8192`, `--enforce-eager`, AITER, and Triton MoE.
Only MI300X and MI355X are marked verified. Generated `public/` files are not
committed.

This contribution was developed with OpenAI Codex assistance.

### DIFF
--- a/models/zai-org/GLM-5.3-Flash.yaml
+++ b/models/zai-org/GLM-5.3-Flash.yaml
@@ -13,6 +13,7 @@ meta:
     h100: verified
     b200: verified
     gb200: verified
+    mi300x: verified
     mi355x: verified
 
 model:
@@ -21,12 +22,12 @@ model:
   nightly_required: true
   docker_image:
     nvidia: "vllm/vllm-openai:glm53-flash"
-    amd: "vllm/vllm-openai-rocm:glm53-flash"
+    amd: "vllm/vllm-openai-rocm:nightly"
   install:
     docker:
-      note: "Use the dedicated GLM-5.3-Flash docker until support lands in the standard vLLM image."
+      note: "Use the dedicated NVIDIA image or a ROCm nightly that includes the AMD integration until support lands in a tagged vLLM image."
     pypi:
-      note: "Use the dedicated GLM-5.3-Flash docker until support lands in the standard vLLM image."
+      note: "Use the appropriate Docker image until GLM-5.3-Flash support lands in a tagged vLLM release."
   architecture: moe
   parameter_count: "321B"
   active_parameters: "18B"
@@ -89,10 +90,18 @@ hardware_overrides:
       - "fp8"
   amd:
     extra_args:
+      - "--trust-remote-code"
+      - "--max-model-len"
+      - "8192"
       - "--max-num-seqs"
-      - "512"
-      - "--attention-backend"
-      - "ROCM_AITER_MLA_SPARSE"
+      - "64"
+      - "--gpu-memory-utilization"
+      - "0.85"
+      - "--enforce-eager"
+      - "--generation-config"
+      - "vllm"
+      - "--moe-backend"
+      - "triton"
     extra_env:
       VLLM_ROCM_USE_AITER: "1"
 
@@ -127,16 +136,24 @@ guide: |
   through 8 of 288 experts, and supports image and video inputs. The checkpoint declares
   a maximum context length of 1,048,576 tokens and includes one MTP draft layer.
 
-  The implementation supports NVIDIA Hopper and newer GPUs, and AMD Instinct
-  gfx950 via ROCm. It can scale with
-  tensor, pipeline, expert, or data+expert parallelism.
+  The current upstream vLLM implementation supports NVIDIA Hopper and newer GPUs.
+  AMD Instinct MI300X, MI325X, and MI355X support is provided by the pending
+  [ROCm integration](https://github.com/ZJY0516/vllm/pull/5). MI325X uses the
+  same gfx942 execution path as MI300X, but has not been independently
+  runtime-validated.
+
+  The model can scale with tensor, pipeline, expert, or data+expert parallelism.
 
   ## Prerequisites
 
   - **Weights:** about 306 GiB for the default native FP8 checkpoint before runtime
     and KV-cache overhead; the BF16 variant requires roughly twice the weight memory
-  - **vLLM:** use docker before the integration is included in the public repo
-  - **FlashInfer:** 0.6.17 or newer is required for NoPE sparse MLA
+  - **vLLM:** use the dedicated NVIDIA image, or a ROCm nightly built after the
+    AMD integration lands, until support is included in a tagged release
+  - **NVIDIA:** FlashInfer 0.6.17 or newer is required for NoPE sparse MLA
+  - **AMD ROCm:** use 8x MI300X / MI325X / MI355X with AITER enabled. The recipe's
+    8K eager-mode configuration is the validated starting point; increase the context
+    length or enable graph capture only after validating the target deployment.
 
   The default model ID, `zai-org/GLM-5.3-Flash`, is the FP8 checkpoint. Select the BF16
   variant to serve `zai-org/GLM-5.3-Flash-BF16` instead.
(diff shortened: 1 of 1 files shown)


ziliangpeng:
I just filed another PR to state glm 5.3 flash doesn't work on MI325 yet.

thanks for this!

stefanskiasan:
**Throughput numbers for the gap this PR explicitly leaves open — measured on 8× MI350X (gfx950) with your validated configuration.**

The PR notes it "does not claim ... throughput validation", and sgl-project/sglang#36608 says the same ("No throughput or latency benchmark was run"). As far as I can find, no throughput figure for GLM-5.3-Flash on any AMD part has been published in either project, so here is one.

**Configuration** — deliberately identical to the baseline you validated accuracy against: TP8, `--max-model-len 8192`, eager, `VLLM_ROCM_USE_AITER=1`, Triton MoE (`VLLM_ROCM_USE_AITER_MOE=0`), `--kv-cache-dtype bfloat16`, `--reasoning-parser glm45 --tool-call-parser glm47`, `--max-num-seqs 64`, `--gpu-memory-utilization 0.85`.

**Hardware/software**: 8× MI350X (gfx950), ROCm 7.2.3, Triton 3.7.1, `zai-org/GLM-5.3-Flash`, `ZJY0516/vllm@glm-release` including ZJY0516/vllm#5, plus the AITER sparse-MLA files from ROCm/aiter#4919 mounted over the image tree.

**Workload**: 8 fixed short English technical prompts (round-robin), `temperature=0.0`, `max_tokens=1024`, concurrency swept. Generation rate is `completion_tokens / wall_clock` across the whole batch.

| N | completed | `stop` | wall | generated | prompt | answers/s | median latency | tokens/answer |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 1/1 | 1 | 34.7 s | **12 tok/s** | 1 | 0.03 | 34.7 s | 404 |
| 8 | 8/8 | 3 | 85.3 s | **82 tok/s** | 2 | 0.09 | 85.3 s | 874 |
| 32 | 32/32 | 10 | 86.6 s | **327 tok/s** | 9 | 0.37 | 86.6 s | 886 |
| 64 | 64/64 | 17 | 87.4 s | **659 tok/s** | 17 | 0.73 | 87.3 s | 899 |
| 128 | 128/128 | 44 | 169.9 s | **670 tok/s** | 17 | 0.75 | 85.3 s | 889 |

Zero errors, zero empty generations, zero faults across all runs.

Two caveats worth stating plainly:

1. **Most responses hit `max_tokens`, not `stop`** (17/64 at N=64). This is the reasoning-template behaviour you already document — the checkpoint ignores `enable_thinking=false` and defaults to maximum reasoning effort — so these are thinking-heavy generations. The token rate is therefore a fair measure of raw decode throughput but *not* of finished answers per second, which is far lower (0.73/s at N=64).
2. **Saturation is at `--max-num-seqs 64`**: 659 → 670 tok/s from N=64 to N=128 is flat, and per-request median latency stays at ~85 s while wall-clock doubles, i.e. the extra 64 requests simply queue. A higher `--max-num-seqs` would be needed to find the real ceiling; I did not sweep it.

For scale on the same 8 GPUs, DeepSeek-V4-Flash (284B/13B active, fp8 KV, CUDA graphs, DSpark-3 speculative decoding) reaches 4.3–5.5 k tok/s on this box. The gap is mostly the three constraints this configuration carries — eager, BF16 KV, Triton MoE — rather than anything intrinsic to the model, so I would read 660 tok/s as a floor for gfx950 rather than a representative figure.

Happy to re-run any of this, sweep `--max-num-seqs`, or produce a GSM8K run for comparison against your 0.9750 on MI355X.

stefanskiasan:
**Follow-up to my numbers above: two of the settings this recipe recommends are now leaving a lot on the table on MI350X, and speculative decoding works.**

Same node as before — 8× MI350X (gfx950), ROCm 7.2.3, TP8, BF16 KV, `--max-model-len 131072 --max-num-seqs 256`, decode CUDA graphs on, 8 fixed prompts, `temperature=0.0`, `max_tokens=1024`.

| N | this recipe's settings | + AITER fMoE | **+ MTP** | total |
|---:|---:|---:|---:|---:|
| 64 | 2,328 | 2,632 | **3,760** | +62 % |
| 256 | 2,766 | 3,428 | **5,134** | +86 % |
| 512 | 3,034 | 3,981 | **5,827** | +92 % |

**1. `VLLM_ROCM_USE_AITER_MOE=0`** — correct as written today, because stock AITER faults on the 288-expert/top-8 FP8 shape (I filed that as ROCm/aiter#5049). But the fix candidate on that issue (`610de6deb`) is worth +31 % at N=512, and I validated it this morning at 36/36 on a long-context tool-calling probe against Triton's 34/36, zero faults. Once it lands, this line should flip. Worth noting that the *first* proposed fix for the same issue booted cleanly and was equally fast but degraded output badly past ~25k tokens (1/9 on the same probe) — throughput and no-crash were not sufficient evidence, which is why I would not recommend flipping the flag before that specific commit is in.

**2. Speculative decoding is not mentioned, and the checkpoint ships for it.** GLM-5.3-Flash carries one MTP layer (`num_nextn_predict_layers = 1`, layer 45 with `eh_proj`/`enorm`/`hnorm`). Enabling it is the single largest win here:

```
--speculative-config '{"method":"mtp","num_speculative_tokens":1}'
```

Draft acceptance rate **84.7 %**, acceptance length 1.85 of a maximum 2, +46 % on top of the MoE fix at N=512.

It does not work out of the box on ROCm: the rope-free decode dispatch falls through to the AITER asm kernel (`mla_dec_stage1_bf16_a16w16_subQ16_mqa16`) once the draft step makes `max_query_len > 1`, and that faults on all ranks during memory profiling. Details and the one-line selector fix are in vllm-project/vllm#53943. If this recipe adds an MTP line for ROCm, it needs that fix or a note.

**3. fp8 KV — tested and not recommended here.** Slower than BF16 (2,717 vs 2,766 at N=256 in the pre-MoE-fix arm), less accurate at 100k, and it buys no capacity worth having: 31.5M vs 32.3M cache tokens, because only 11 of 45 layers are MLA. The 34 KDA layers dominate the footprint and are unaffected by the KV dtype.

Correctness alongside the throughput, since none of the above is useful if the output degrades: 12/12 correct tool calls with MTP across two runs at 167 / 27k / 55k / 82k / 109k / 123k prompt tokens, and the full 1M window still holds at 972,502 prompt tokens with a correct tool call.

Happy to contribute an MI350X section with the full command line if that is in scope for this recipe — say the word and I will open a PR against it rather than leaving this in a comment.

PR #861, opened 2026-08-27
Kimi-K3: set --prefix-match-unit 128 on the Blackwell profile

NickLucche:
## What

Adds `--prefix-match-unit 128` to the Kimi-K3 Blackwell hardware profile (where `--enable-prefix-caching` is already enabled), plus a guide note.

## Why

K3 is a hybrid model: its MLA attention layers and KDA (Mamba-like) layers form two KV-cache groups under the hybrid KV-cache manager. With prefix caching on, vLLM pads the attention block size up to match the Mamba state page, so **both groups resolve to a 1536-token block**. The default prefix-cache hit boundary is therefore every 1536 tokens — very coarse.

`128` divides 1536 and aligns with the MLA kernel's native block boundary (`kernel_block_alignment_size`), giving **12x finer prefix-hit granularity** (a hit boundary every 128 tokens instead of 1536). The inflation formula in `vllm/platforms/interface.py` always yields a multiple of 128 across TP sizes, so 128 is valid under every tensor-parallel layout. The flag is a no-op where prefix caching is off, so it is scoped to the Blackwell hardware override (the only profile that enables `--enable-prefix-caching`).

## Scope

- `models/moonshotai/Kimi-K3.yaml` only — Blackwell `hardware_overrides` block + guide note + `date_updated` bump.
- Hopper / AMD / Ascend profiles are untouched (they don't enable prefix caching, so the flag would be a no-op there).

## Validation

- `node scripts/build-recipes-api.mjs` → `✓ JSON API: 178 models` (clean parse).
- Confirmed the flag lands in the Blackwell `recommended_command` (`public/moonshotai/Kimi-K3.json`) and does **not** leak into the Hopper/AMD per-hw renderings.

## Notes

- Not duplicating an existing PR: no recipe or vLLM doc sets `--prefix-match-unit` for K3 today.
- AI assistance was used to investigate the block-size resolution; the change and rationale were reviewed end-to-end.

Co-authored-by: Mistral Vibe <vibe@mistral.ai>

### DIFF
--- a/models/moonshotai/Kimi-K3.yaml
+++ b/models/moonshotai/Kimi-K3.yaml
@@ -254,10 +254,11 @@ hardware_overrides:
       - '{"use_prefill_query_quantization":true,"mla_prefill_backend":"flashinfer"}'
       - "--enable-prefix-caching"
       # K3's two KV-cache groups (MLA attention + KDA Mamba) both resolve to a
-      # 1536-token block under the hybrid manager, so the default prefix-cache
-      # hit boundary is every 1536 tokens. 128 divides 1536 (the inflation formula
-      # always yields a multiple of 128) and matches the MLA kernel block
-      # alignment, giving 12x finer prefix hits. No-op without prefix caching.
+      # single block size under the hybrid manager: vLLM pads the attention
+      # block up to match the Mamba state page, so the default prefix-cache hit
+      # boundary is that (large) block size. The inflation formula always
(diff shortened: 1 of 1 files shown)


NickLucche:
cc @ZJY0516 happy to wait out on this one, but I think it would help to let users be aware of this feature :)

esmeetu:
LGTM.

PR #891, opened 2026-08-30
Fix stale version note in Qwen3.8-27B recipe

namgyu-youn:
## Summary
`Qwen3.8-27B.yaml`: the DFlash2 speculative-decoding mode said its dependency (https://github.com/vllm-project/vllm/pull/52816) was "not in a stable release yet." It's recently released on [v0.28.0](https://github.com/vllm-project/vllm/releases/tag/v0.28.0) (2026-08-26), so update to say "vLLM >=0.28.0" instead.

## Test plan
- `git show v0.28.0:tests/v1/spec_decode/test_dflash2.py` confirms DFlash2 (PR #52816) is present in the v0.28.0 tag.

### DIFF
--- a/models/Qwen/Qwen3.8-27B.yaml
+++ b/models/Qwen/Qwen3.8-27B.yaml
@@ -111,7 +111,7 @@ features:
           - '{"method":"qwen3_5_mtp","num_speculative_tokens":3,"enforce_eager":true}'
       dflash:
         label: "DFlash2"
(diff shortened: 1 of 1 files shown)


namgyu-youn:
Correction: the Flash-Next compile-config change in this PR was based on a flawed test — I checked the dotted `-cc.mode`/`-cc.cudagraph_mode` syntax against plain Python `argparse.ArgumentParser` instead of vLLM's actual `FlexibleArgumentParser` (which `vllm serve` uses and which has custom preprocessing for dotted CLI args). Re-tested against the real parser plus the actual `CompilationConfig` validation, and the original syntax works correctly. Reverted that part in the latest commit — this PR now only contains the DFlash2 version-note fix, which is unaffected and independently verified. Apologies for the noise.

esmeetu:
Thanks!

PR #897, opened 2026-08-31
Remove plamo-2-translate recipe (Plamo2 architecture removed from vLLM)

namgyu-youn:
## Summary
vLLM v0.27.0 dropped `pfnet/plamo-2-translate` (uses the `Plamo2ForCausalLM` architecture) support entirely — https://github.com/vllm-project/vllm/pull/49729

The model file is deleted upstream and the architecture is now hard-rejected via `_PREVIOUSLY_SUPPORTED_MODELS`, with no successor checkpoint or out-of-tree plugin available. Since serving this recipe now fails on any current vLLM install with no working fallback, the recipe is removed rather than pinned to an increasingly old vLLM version.

## Test plan
- `node scripts/build-recipes-api.mjs` builds cleanly (180 → 179 models, no new errors)


esmeetu:
lgtm

PR #902, opened 2026-08-31
Update vLLM configuration in gpt-oss-120b.yaml (B200)

aas008:
Added comments for vLLM configuration changes and updated extra_args for Blackwell which were depreciated in [PR#43148 ](https://github.com/vllm-project/vllm/pull/43148)

### DIFF
--- a/models/openai/gpt-oss-120b.yaml
+++ b/models/openai/gpt-oss-120b.yaml
@@ -64,6 +64,13 @@ compatible_strategies:
   - multi_node_dep
 
 hardware_overrides:
+  blackwell:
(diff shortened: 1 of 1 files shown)


sharonh24:
This issue also relates to [#37128](https://github.com/vllm-project/vllm/pull/37128), where gpt_oss_120b was kept in BF16 due to minor FP8 accuracy drops below threshold. Consequently, gpt_oss misses out on recent vLLM MoE and prefill/decode performance gains.

Could the accuracy tests from [#37128](https://github.com/vllm-project/vllm/pull/37128) be re-run to check if the FP8 accuracy regression still holds on newer vLLM versions?

mgoin (approved):

PR #913, opened 2026-09-02
Add Atlas 800I A2 W8A8 recipe for Qwen3-30B-A3B

chaosv598:
## Background

Qwen3-30B-A3B has an existing Intel Xeon 6 BF16 recipe. This adds the separately validated Huawei Atlas 800I A2 (8× Ascend 910B) single-node W8A8 path while preserving the CPU/BF16 default behavior.

## Main changes

1. Add the restricted `ascend_910b_8x` hardware profile for Atlas 800I A2 (8×64 GB Ascend 910B).
2. Add an `ascend_w8a8` variant using the public ModelScope `Eco-Tech/Qwen3-30B-A3B-w8a8` checkpoint. It is scoped to the new profile and pins TP4, `--quantization ascend`, expert parallelism, the multiprocessing executor, and the validated runtime environment.
3. Pin the exact A2 variant image to `quay.io/ascend/vllm-ascend:v0.23.0rc1` and document a reproducible launch plus OpenAI-compatible smoke checks.

## Scope and command behavior

- The A2 profile is `restricted`, so it appears only on recipes that explicitly record validation.
- TP4 is variant- and profile-specific; it does not change the existing BF16 default or Xeon 6 configuration.
- The W8A8 checkpoint is ModelScope-hosted. The guide calls out `VLLM_USE_MODELSCOPE=True` and links the exact checkpoint.

## Validation

- Parsed both modified YAML files and asserted the taxonomy/recipe relationship, A2-only hardware allowlist, variant-specific TP4, and unchanged BF16 default.
- The serving configuration was validated end-to-end on Atlas 800I A2: server readiness, `/v1/models`, and chat-completions smoke request.

### DIFF
--- a/models/Qwen/Qwen3-30B-A3B.yaml
+++ b/models/Qwen/Qwen3-30B-A3B.yaml
@@ -2,17 +2,18 @@ meta:
   title: "Qwen3-30B-A3B"
   slug: "qwen3-30b-a3b"
   provider: "Qwen"
-  description: "Qwen3 MoE model with 30.5B total and 3.3B active parameters, validated for BF16 serving on Intel Xeon 6."
+  description: "Qwen3 MoE model with 30.5B total and 3.3B active parameters, validated for BF16 serving on Intel Xeon 6 and W8A8 serving on Huawei Atlas 800I A2."
   date_added: 2026-08-14
-  date_updated: 2026-08-14
+  date_updated: 2026-09-02
   difficulty: intermediate
   tasks:
     - text
-  performance_headline: "Validated on Intel Xeon 6 (GNR) with BF16 serving"
+  performance_headline: "Validated on Intel Xeon 6 (GNR) with BF16 serving and Atlas 800I A2 with W8A8 serving"
   related_recipes:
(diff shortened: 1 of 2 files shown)


models/Qwen/Qwen3-30B-A3B.yaml
```diff
@@ -44,6 +45,29 @@ variants:
...
+    supported_hardware: [ascend_910b_8x]
+    tp:
+      ascend_910b_8x: 4
+    extra_args:
+      - "--quantization"
+      - "ascend"
+      - "--enable-expert-parallel"
+      - "--distributed-executor-backend"
+      - "mp"
+    extra_env:
```
esmeetu, line 62: nit: `VLLM_USE_MODELSCOPE=True`

models/Qwen/Qwen3-30B-A3B.yaml
```diff
@@ -110,6 +134,35 @@ guide: |
       --port 8000
   ```
 
+  ## Huawei Atlas 800I A2 (8× Ascend 910B)
```
esmeetu, line 137: I think we can drop these since we already cover something similar above.

PR #922, opened 2026-09-03
fix incorrect parser in NVIDIA Nemotron recipe

namgyu-youn:
## Summary
NVIDIA Nemotron recipe is using `--tool-call-parser-plugin`. However, `--tool-call-parser-plugin` is wrong parser; vLLM uses `--tool-parser-plugin` instead.

## Test Plan
- `node scripts/build-recipes-api.mjs` — YAML parses, JSON API regenerates cleanly (187 models, no errors).


esmeetu:
lgtm!

PR #929, opened 2026-09-05
Add DFlash2 spec decoding mode to GLM-5.3-Flash

yiminghub2024:
## Summary
- Add a DFlash2 mode to GLM-5.3-Flash `spec_decoding`, keeping built-in MTP as the default.
- Selecting DFlash2 emits `--speculative-config '{"method":"dflash","model":"incoai/GLM-5.3-Flash-DFlash2","num_speculative_tokens":7}'` (block size 8 minus one).
- AMD stays on MTP: the ROCm GLM-5.3-Flash image does not support this speculative path.
- Guide text is unchanged; only the command-builder config is updated.

## Test plan
tested on 8*b300 ,its work fine~


gaby:
@yiminghub2024 Which version or tag of vLLM you used for testing this? I tried last week and it fails to start.

yiminghub2024:
docker pull vllm/vllm-openai:glm53-flash or vllm/vllm-openai:nightly , i forget ,you can try there two , @gaby

gaby:
@yiminghub2024 I don't know how you tested this, because it doesn't work. The model won't even start with this error:

```
RuntimeError: Model does not support EAGLE3 interface
```

I just tested this with the latest nightly.

Related:
- https://github.com/vllm-project/vllm/pull/55423
- https://github.com/vllm-project/vllm/pull/55682

yiminghub2024:
@gaby  yes ,i tested it you are right ,sorry ,i mistake start pr for glm-5.3-flash ,i tested it on b300 on glm-5.3 , glm-5.3 can start with dlfash2 fine.    thanks for your discover and point out  ， we will wait for your  https://github.com/vllm-project/vllm/pull/55682 merge .

PR #931, opened 2026-09-05
Embed vLLM benchmark charts on vLLM recipes

functionstackx:
# Human

InferenceX closely follows recipes.vllm.ai recipes and measures the performance of recipes.vllm.ai . This embeds inferencex Agentic workload graphs into vLLM recipes 

### minimax m3 (dark)

https://vllm-recipes-3g6bfk6gh-inferact-inc.vercel.app/MiniMaxAI/MiniMax-M3

image

## minimax m3 (light)

image

## kimi k3 (dark)

https://vllm-recipes-3g6bfk6gh-inferact-inc.vercel.app/moonshotai/Kimi-K3

image

## kimi k3 (light)

image

# AI Written Below
## Summary

Prototype: embed live InferenceX benchmark charts on recipe pages, scoped to vLLM results only.

Recipes opt in with one key in their YAML:

```yaml
inferencex:
  model: deepseek-v4   # InferenceX compare slug
```

When present, the recipe page gets a "Benchmarks (InferenceX)" accordion above the Guide with an iframe to `<INFERENCEX_BASE_URL>/embed/model/<slug>?framework=vllm&theme=<vllm-light|vllm-dark>`. The `framework=vllm` lock is applied inside InferenceX, so SGLang, ATOM, and TRT-LLM configs never appear in the plot, legend, filter dialog, or config changelog. The embed footer says "Showing vLLM results only" and links to the full dashboard.

Enabled for:

- `deepseek-ai/DeepSeek-V4-Pro` -> `deepseek-v4`
- `moonshotai/Kimi-K3` -> `kimi-k3`
- `MiniMaxAI/MiniMax-M3` -> `minimax-m3`

## Changes

- `src/lib/inferencex.js`: `resolveInferencex(meta)`, `inferencexEmbedUrl(cfg, theme)`, `inferencexModelUrl(cfg)`. Base URL defaults to `https://inferencex.semianalysis.com`, overridable with `NEXT_PUBLIC_INFERENCEX_URL` for previews against a local or staging InferenceX.
- `src/components/recipes/InferenceXEmbed.jsx`: client component. Watches the `dark` class on `<html>` and reloads the iframe with the matching `theme`. Listens for `inferencex:embed-resize` postMessages (origin and source checked) and sets the iframe height so there is no inner scrollbar. `sandbox="allow-scripts allow-same-origin allow-popups"`, `loading="lazy"`, min height 560px.
- `src/app/[org]/[repo]/page.js`: renders the accordion when `resolveInferencex(recipe.meta)` returns a config.
- Three model YAMLs: added the `inferencex` block.
- `CLAUDE.md`: documented the new meta key.

`scripts/build-recipes-api.mjs` passes and the JSON API carries `meta.inferencex`. Full `pnpm build` and `next start` verified locally; screenshots (light and dark) are in the Slack thread.

## Theme and chrome

The iframe requests InferenceX's `vllm-light` / `vllm-dark` skin rather than the stock theme. The skin reuses this site's slate palette, Inter and JetBrains Mono, and the vLLM yellow/blue accents, and hides everything a static embed should not carry: axis metric pickers, compare history, run dates, config changelog, quick filters, Optimal Only, gradient labels, and the legend switches. What remains is the chart, a read-only legend, and the source/updated caption. The recipe-side wrapper border was dropped since the embed's own card now provides it.

## Depends on

SemiAnalysisAI/InferenceX-app#1008 (merged, live in production) adds the `/embed/model/<slug>` route. The iframe points at `https://inferencex.semianalysis.com` by default, e.g. https://inferencex.semianalysis.com/embed/model/kimi-k3?framework=vllm&theme=vllm-light. `NEXT_PUBLIC_INFERENCEX_URL` is only needed to preview against a local or staging InferenceX.

Update (1e34095, 3212d2d): the "Open the full interactive dashboard on InferenceX" link now sits between the caption and the iframe and points at `/inference/<slug>`. The embed keeps the dashboard's 600px plot and hides the zoom hint and the request-stats footer line (InferenceX-app #1014, #1016).


src/components/recipes/InferenceXEmbed.jsx
```diff
@@ -0,0 +1,122 @@
...
+      if (event.origin !== embedOrigin) return;
+      if (event.source !== frameRef.current?.contentWindow) return;
+      const data = event.data;
+      if (!data || data.type !== RESIZE_MESSAGE_TYPE) return;
+      if (typeof data.height !== "number" || !Number.isFinite(data.height)) return;
+      setHeight(Math.max(MIN_HEIGHT, Math.ceil(data.height)));
+    };
+    window.addEventListener("message", onMessage);
+    return () => window.removeEventListener("message", onMessage);
+  }, []);
```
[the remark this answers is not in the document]
functionstackx: Fixed in f538ae3. Rather than a try/catch, the origin is now resolved against the page URL (`new URL(INFERENCEX_BASE_URL, window.location.href).origin`), which handles absolute and relative bases alike and yields the page origin for a relative proxy path, so the postMessage origin check keeps working.

esmeetu (commented):
Thanks for this — the embed is a nice addition and the overall shape (per-recipe opt-in, vLLM-locked chart, themed skin) is sound. Nothing here is a hard blocker, but the iframe component needs some hardening before it ships on `defaultOpen` pages, and I'd like to discuss where the InferenceX mapping lives. Inline comments below; a few points that fall outside the diff:

**Schema placement: `meta.inferencex` vs. a standalone mapping table.** I'd lean towards a top-level `inferencex.yaml` keyed by HF id (`moonshotai/Kimi-K3: kimi-k3`) rather than a `meta` key, for three reasons:
1. *Ownership.* The slug is InferenceX's namespace, not a recipe attribute, and it isn't derivable from the HF id. When InferenceX adds or renames a batch of models, one table edit beats touching N recipe YAMLs (and avoids tempting a `date_updated` bump for what isn't a material recipe change).
2. *JSON API shape.* `meta` is published verbatim, so every promoted quantized variant (`nvidia/DeepSeek-V4-Pro-NVFP4.json`, `Eco-Tech/Kimi-K3-w4a8.json`, `amd/MiniMax-M3-MXFP4.json`, …) currently inherits the parent's slug even though InferenceX benchmarked a specific checkpoint. A table keyed by HF id gives exact variant granularity and lets the build script emit one resolved `{ model, url }` block instead of an opaque slug.
3. *Room to grow.* Base URL / `framework=vllm` / skin are module constants today; a table is a natural home if a second benchmark source ever shows up.

The one cost is stale keys, so the table needs a check in `build-recipes-api.mjs` that every key resolves to a recipe or promoted-variant hf_id. That check is needed either way — today nothing at build time inspects `meta.inferencex` at all. If you'd rather keep `meta`, that's consistent with the `platforms` precedent and fine by me, but then please (a) drop the undocumented string shorthand so the API shape is stable, (b) drop or type-check `scenario`, and (c) normalise through `resolveInferencex` in the build script and publish a resolved URL.

**JSON API** (`scripts/build-recipes-api.mjs`): as above — `meta.inferencex` passes through HEAD_KEYS untouched, so consumers get a bare slug with no URL and have to reverse-engineer `https://inferencex.semianalysis.com/inference/<slug>` from `src/lib/inferencex.js`. That module has no Next deps, so the script can import it.

**Docs left stale**: the new key is only in CLAUDE.md. `.claude/skills/add-recipe/SKILL.md` (the `meta:` template, and the line saying narrative claims live in the Guide accordion only) and CONTRIBUTING.md's schema block don't mention it, and CLAUDE.md's Architecture entry for the recipe page still says header + CommandBuilder + guide. The next person authoring a recipe for a benchmarked model won't learn the opt-in exists.

src/components/recipes/InferenceXEmbed.jsx
```diff
@@ -0,0 +1,124 @@
...
+          />
+        )}
+        {src && (
+          <iframe
+            ref={frameRef}
+            key={src}
+            src={src}
+            title={`InferenceX benchmarks for ${title} on vLLM`}
+            loading="lazy"
+            onLoad={() => setLoadedSrc(src)}
```
esmeetu, line 115: **No error / timeout / fallback state.** The skeleton is cleared only by `onLoad`. With an ad blocker, a corporate firewall, or InferenceX returning 5xx, Chromium/Firefox still fire `load` for the error document, so `loadedSrc === src` hides the skeleton and a blank 560px card sits above the Guide in a `defaultOpen` section. A stalled connection never fires `load` and the skeleton pulses forever.

Suggest an `onError` handler plus a timer (~8–10s) that swaps the frame for a short "Chart unavailable — open on InferenceX" link using `inferencexModelUrl(config)`.

src/components/recipes/InferenceXEmbed.jsx
```diff
@@ -0,0 +1,124 @@
...
+    // Resolve against the page so a relative NEXT_PUBLIC_INFERENCEX_URL (a
+    // same-origin proxy in previews) yields the page origin instead of throwing.
+    const embedOrigin = new URL(INFERENCEX_BASE_URL, window.location.href).origin;
+    const onMessage = (event) => {
+      if (event.origin !== embedOrigin) return;
+      if (event.source !== frameRef.current?.contentWindow) return;
+      const data = event.data;
+      if (!data || data.type !== RESIZE_MESSAGE_TYPE) return;
+      if (typeof data.height !== "number" || !Number.isFinite(data.height)) return;
+      setHeight(Math.max(MIN_HEIGHT, Math.ceil(data.height)));
```
esmeetu, line 55: **Height is clamped only at the bottom.** The wrapper is `style={{ height }}` and the iframe is `h-full`, so the embed's viewport equals the last reported height. If the embed measures `body.scrollHeight` while its body is 100%-tall plus margin (the classic iframe-resizer feedback loop), each report exceeds the current height and the card diverges to thousands of px; a single stray `height: 1e7` yields a 10,000,000px card since only origin/source/type/isFinite are checked.

Suggest a `MAX_HEIGHT` cap and ignoring deltas below a few px.

src/components/recipes/InferenceXEmbed.jsx
```diff
@@ -0,0 +1,124 @@
...
+      <a
+        href={inferencexModelUrl(config)}
+        target="_blank"
+        rel="noopener noreferrer"
+        className="inline-flex items-center gap-1.5 text-base font-medium text-vllm-blue hover:underline transition-colors"
+      >
+        Open the full interactive dashboard on InferenceX
+        <ExternalLink size={16} />
+      </a>
+      <div className="relative rounded-xl overflow-hidden bg-background" style={{ height }}>
```
esmeetu, line 101: **The height contract has no fallback.** Every guard in `onMessage` is a bare `return`, so if InferenceX renames the message type, errors before posting, or serves the embed from an origin other than `INFERENCEX_BASE_URL`, the frame stays pinned at 560px with an internal scrollbar (contradicting the docblock) and nothing is logged. Conversely a late `height: 900` report pushes the Guide ~340px after a visitor has already scrolled to a `#heading` (rehype-slug ids are deep-linkable), and a short "no vLLM runs yet" embed leaves dead `bg-background` space under the permanent `MIN_HEIGHT` floor.

At minimum a `console.warn` in dev on rejected messages; ideally an `aspect-ratio` fallback so the frame is usable even when no resize ever arrives.

src/components/recipes/InferenceXEmbed.jsx
```diff
@@ -0,0 +1,124 @@
...
+        {src && (
+          <iframe
+            ref={frameRef}
+            key={src}
+            src={src}
+            title={`InferenceX benchmarks for ${title} on vLLM`}
+            loading="lazy"
+            onLoad={() => setLoadedSrc(src)}
+            className="block h-full w-full border-0"
+            sandbox="allow-scripts allow-same-origin allow-popups allow-popups-to-escape-sandbox"
```
esmeetu, line 117: **`sandbox` is a no-op in the same-origin proxy mode the code explicitly supports.** With `NEXT_PUBLIC_INFERENCEX_URL=/ix` (the relative case the comment at L46 describes), the frame is same-origin with both `allow-scripts` and `allow-same-origin`, which per the HTML spec lets the framed document remove its own sandbox attribute, reload, and script `window.parent` (localStorage hardware/theme prefs, the command block).

The reverse footgun also exists: if a maintainer later "tightens" this by dropping `allow-same-origin`, `event.origin` becomes the literal string `"null"`, every resize message is rejected by the origin check, and the frame is silently stuck at 560px. Worth deciding which mode is actually supported and documenting the coupling between this attribute and the origin guard.

src/components/recipes/InferenceXEmbed.jsx
```diff
@@ -0,0 +1,124 @@
...
+        <a
+          href={inferencexModelUrl(config)}
+          target="_blank"
+          rel="noopener noreferrer"
+          className="text-vllm-blue hover:underline"
+        >
+          InferenceX
+        </a>{" "}
+        following{" "}
+        <a href="https://recipes.vllm.ai" className="text-vllm-blue hover:underline">
```
esmeetu, line 88: Hard-coded `href="https://recipes.vllm.ai"` is the only `href` in `src/` that bypasses the site's canonical-URL handling, so on Vercel previews and local dev this navigates out of the current deployment to production.

Use `href="/"` like `layout.js` and `not-found.js` do. Don't import `siteUrl` into this `"use client"` file — `VERCEL_ENV`/`VERCEL_URL` aren't `NEXT_PUBLIC`, so client and server would disagree and you'd get a hydration mismatch. If you need the absolute URL, pass it as a prop from `page.js`.

src/app/[org]/[repo]/page.js
```diff
@@ -220,6 +224,12 @@ export default async function RecipePage({ params, searchParams }) {
 
       {/* ── Reference sections ── */}
       <section className="space-y-2">
+        {inferencex && (
+          <Accordion title="vLLM Performance - InferenceX Agentic Workloads" defaultOpen>
```
Comment (esmeetu) [inline src/app/[org]/[repo]/page.js:228]: **Every visit pays for the embed.** The accordion is `defaultOpen` and the iframe mounts unconditionally after hydration, so each visit to the three opted-in pages downloads and executes roughly 1.7 MB raw (~560 KB compressed: HTML + 28 JS chunks + CSS + fonts + data XHRs) of third-party assets, even for a visitor who only wants the `vllm serve` command.

`loading="lazy"` doesn't help here — the frame sits directly under CommandBuilder, well inside Chrome's distance threshold — and `defaultOpen={false}` wouldn't either, since `<details>` keeps children in the DOM and Chrome eagerly loads hidden lazy iframes. Suggest an IntersectionObserver-gated mount or a click-to-load facade that conditionally renders `<InferenceXEmbed>`.

src/lib/inferencex.js
```diff
@@ -0,0 +1,48 @@
+// InferenceX (inferencex.semianalysis.com) embed helpers.
+//
+// A recipe opts in with `meta.inferencex: { model: <inferencex-model-slug> }`.
+// The slug is the model's path on InferenceX (`/inference/<slug>`), which is not
+// derivable from the HF id (DeepSeek-V4-Pro → `deepseek-v4`), so it is
+// declared per recipe. The embed is locked to vLLM: this is the vLLM recipes
+// site, so the chart only shows vLLM configurations regardless of what other
+// engines InferenceX benchmarks.
+
+export const INFERENCEX_BASE_URL = (
```
esmeetu, line 10: `NEXT_PUBLIC_INFERENCEX_URL` is never validated, and relative-base support is only half-implemented: the origin check in the component tolerates a relative value, but `inferencexModelUrl` / `inferencexAgentxUrl` then emit relative links on the recipes host.

- `NEXT_PUBLIC_INFERENCEX_URL=/ix-proxy` → embed loads, but the caption/dashboard anchors become `href="/ix-proxy/inference/kimi-k3"` and `/ix-proxy/agentx` on recipes.vllm.ai, which 404 unless the proxy mirrors those pages (`next.config.mjs` defines no rewrite).
- `NEXT_PUBLIC_INFERENCEX_URL=localhost:3000` → `new URL('localhost:3000', href).origin === 'null'`, so every resize is dropped, and `src="localhost:3000/embed/…"` renders a blank frame with no hint that the env var is malformed.

A one-time `new URL()` at module load (throw on failure, export `INFERENCEX_ORIGIN`) would catch both.

src/lib/inferencex.js
```diff
@@ -0,0 +1,48 @@
...
+// part of the page rather than an InferenceX screenshot.
+export const INFERENCEX_SKIN = "vllm";
+
+/** Normalize `meta.inferencex` (string shorthand or object) to `{ model, scenario? }`. */
+export function resolveInferencex(meta) {
+  const raw = meta?.inferencex;
+  if (!raw) return null;
+  const cfg = typeof raw === "string" ? { model: raw } : raw;
+  if (!cfg.model || typeof cfg.model !== "string") return null;
+  return { model: cfg.model, scenario: cfg.scenario || null };
```
esmeetu, line 27: `scenario` is speculative surface: no recipe sets it, CLAUDE.md lists it only as `scenario?` with no vocabulary, nothing confirms InferenceX accepts a `scenario` query param, and unlike `model` it goes untyped into `URLSearchParams.set`. An author guessing `scenario: [agentx, chat]` or `scenario: { id: agentx }` gets `scenario=agentx%2Cchat` / `%5Bobject+Object%5D`, InferenceX serves the default or an empty chart, and no build-time check fires.

Suggest dropping it until a recipe needs it, or gating with `typeof cfg.scenario === "string"` and documenting the allowed values. Same for the string shorthand at L25 — it's accepted here but absent from CLAUDE.md, and it makes the published `meta.inferencex` shape vary between recipes.

models/moonshotai/Kimi-K3.yaml
```diff
@@ -10,6 +10,8 @@ meta:
     - multimodal
     - text
   performance_headline: "Pre-release TP8, TEP16, TP8xPP2, and disaggregated P/D profiles for the 2.8T MXFP4 checkpoint"
+  inferencex:
```
esmeetu, line 13: See the review summary on `meta.inferencex` vs. a standalone `inferencex.yaml` mapping keyed by HF id. Concretely for this recipe: `Eco-Tech/Kimi-K3-w4a8` is promoted as its own JSON and inherits `inferencex: { model: kimi-k3 }` from here, although InferenceX benchmarked the MXFP4 checkpoint. A per-HF-id table would let you opt that variant out (or point it at its own slug) without touching this file.

PR #942, opened 2026-09-09
[Ascend] Offer DeepSeek-V4-Flash-0731 with DSpark on 950PR and 950DT

yiminghub2024:
Offer DeepSeek-V4-Flash-0731 with DSpark on 950PR and 950DT


yiminghub2024:
https://github.com/vllm-project/vllm-ascend/pull/16104

PR #945, opened 2026-09-09
Pin stale nightly recipes to v0.28.0

### LINKED ISSUE #944: [Bug] nightly_required goes stale after a release: 4 recipes force nightly wheels for support already in v0.28.0
## Summary

`nightly_required: true` is a time sensitive claim with no expiry. It is correct when a recipe is authored during the window between an upstream merge and the next release, and it silently becomes wrong the moment that release ships. Nothing re-checks it, so recipes keep telling users to install nightly wheels for support that is already in a stable wheel.

Right now 4 recipes are in that state, and the v0.28.0 cut on 2026-08-26 is what made most of them stale at once.

## Affected recipes

| Recipe | declares | recipe `date_updated` | evidence support is in a stable release |
|---|---|---|---|
| `meta-models/Muse-Glimmer-30B` | `min: nightly`, `nightly_required`, `install.pip: false` | 2026-08-23 | `muse_glimmer.py` present at `v0.28.0`; vllm#51655 merged 08-14 (fixed in #943) |
| `dots-studio/dots3-note-prev` | `min: 0.28.0`, `nightly_required` | 2026-08-14 | `vllm/models/dots3_note/` present at `v0.28.0`, which is the version the recipe itself names |
| `inclusionAI/Ling-3.0-flash` | `min: 0.25.0`, `nightly_required` | 2026-08-31 | `bailing_moe_v3.py` present at `v0.28.0` |
| `inclusionAI/Ling-3.0-tiny` | `min: 0.25.0`, `nightly_required` | 2026-08-31 | same, and the guide's stated need ("a nightly/source build containing Ling 3.0 / Bailing V3") is now satisfied by v0.28.0 |
| `tencent/Hy3` | `min: 0.26.0`, `nightly_required`, pip note says vllm#47433 "aren't in a stable release yet" | 2026-07-06 | vllm#47433 merged 2026-07-06 and is contained in `v0.28.0` (`behind_by: 0`). The recipe was authored the same day that PR merged. |

`dots3-note-prev` is the clearest case: it declares `min_vllm_version: "0.28.0"` and simultaneously forces nightly, but v0.28.0 released on 2026-08-26.

## Secondary finding

The two `inclusionAI/Ling-3.0-*` recipes declare `min_vllm_version: "0.25.0"`, but `bailing_moe_v3.py` does not exist at `v0.25.0`, `v0.26.0`, or `v0.27.0`. It first appears at `v0.28.0`. So the declared floor is lower than reality and should probably read `0.28.0`.

## Deliberately excluded

`thinkingmachines/Inkling` and `thinkingmachines/Inkling-Small` also carry `nightly_required: true` with `min_vllm_version: "0.26.0"`, and `vllm/models/inkling/` is present at `v0.28.0`. I am not listing them as stale because their variants pin `vllm/vllm-openai-rocm:nightly` specifically and the guides reference AITER MXFP4 MoE kernels on ROCm. That may be a genuine nightly dependency on the AMD side rather than base model support. A maintainer with ROCm context should judge those.

`Qwen/Qwen3.8-Flash-Next`, `deepseek-ai/DeepSeek-V4-Flash-Vision-Exp`, `tencent/Hy4-preview`, `zai-org/GLM-5.3-Flash` and `Qwen/Qwen3.8-2.4T-A95B` all correctly require nightly, since they target 0.29.0 which has not shipped.

## Why this matters beyond the version string

Per CONTRIBUTING, `install: { pip: false }` hides the pip tab entirely rather than annotating it, and `nightly_required` swaps the generated pip command for the nightly wheel index. So a stale flag does not just display an old version number, it removes the working install path from the page and sends users to a nightly wheel or a multi GB image they no longer need.

## Suggested remedy

The check is mechanical and does not need a model expert: for every recipe with `nightly_required: true` whose `min_vllm_version` parses to a version less than or equal to the latest release tag, the flag is contradictory by the definition in CONTRIBUTING ("set when `min_vllm_version` hasn't shipped yet"). That is a candidate for a periodic sweep at release time rather than per recipe vigilance. Happy to send that as a script or a CI job if maintainers want it in the repo, and happy to send the per recipe fixes as separate PRs so each vendor's owners can review their own.

#943 fixes the Muse Glimmer instance and shows the shape of the fix.

## Verification method used

For each recipe, rather than relying on merge dates:

```
gh api repos/vllm-project/vllm/compare/<merge_sha>...v0.28.0 --jq .behind_by   # 0 means contained
gh api repos/vllm-project/vllm/contents/<model_module_path>?ref=v0.28.0        # module present at the tag
```

Checking containment plus file presence at the tag avoids the failure mode where something merges before a release but is reverted or moved before the cut.

wqqqqq0720:
## Summary

- remove stale `nightly_required` flags from four recipes whose support shipped in vLLM v0.28.0
- correct the minimum vLLM version for the Ling 3.0 recipes
- update the Hy3 installation guidance now that vLLM #47433 is in a stable release
- refresh `date_updated` for the affected recipes

## Testing

- `git diff --check`

Closes #944

### DIFF
--- a/models/dots-studio/dots3-note-prev.yaml
+++ b/models/dots-studio/dots3-note-prev.yaml
@@ -4,7 +4,7 @@ meta:
   provider: "Dots"
   description: "Multimodal MoE model available in BF16 and native FP8, with hybrid DSA and SWA, 512K context, and MTP speculative decoding."
   date_added: 2026-08-13
-  date_updated: 2026-08-14
+  date_updated: 2026-09-09
   difficulty: advanced
   tasks:
     - multimodal
@@ -23,7 +23,6 @@ meta:
 model:
   model_id: "dots-studio/dots3-note-prev"
   min_vllm_version: "0.28.0"
-  nightly_required: true
   architecture: moe
   parameter_count: "288B"
   active_parameters: "17B"
@@ -128,7 +127,7 @@ guide: |
     memory for the runtime and KV cache; the recipe budgets 692 GB aggregate VRAM.
   - A recent CUDA environment. Native FP8 uses DeepGEMM, while unquantized BF16
     uses the Triton MoE backend.
-  - A vLLM nightly containing native dots3-note-prev support.
+  - vLLM 0.28.0 or newer, which includes native dots3-note-prev support.
 
   ## Launch commands
 
--- a/models/inclusionAI/Ling-3.0-flash.yaml
+++ b/models/inclusionAI/Ling-3.0-flash.yaml
@@ -3,7 +3,7 @@ meta:
   slug: "ling-3-0-flash"
   provider: "inclusionAI"
   description: "Ling-3.0-flash MoE model with BF16, FP8, FP4, and INT4 checkpoints, native MTP, and an external DSpark draft model"
-  date_updated: 2026-08-31
+  date_updated: 2026-09-09
   difficulty: advanced
   tasks:
     - text
@@ -14,8 +14,7 @@ meta:
 
 model:
   model_id: "inclusionAI/Ling-3.0-flash"
-  min_vllm_version: "0.25.0"
-  nightly_required: true
+  min_vllm_version: "0.28.0"
   architecture: moe
   parameter_count: "124B"
   active_parameters: "5.5B"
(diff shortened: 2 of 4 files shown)


esmeetu:
thanks!

PR #952, opened 2026-09-10
[ROCm][DeepSeek-V4.1-Flash] Set the working CUDA graph mode for AMD

JohnQinAMD:
## Summary

vLLM's default `cudagraph_mode` cannot work for `DeepseekV41ForCausalLM` on ROCm, and the failure is not obvious, so the recipe should pin the mode that does work.

Two flags, both verified on 4× MI355X (gfx950), TP4.

### Why `FULL_DECODE_ONLY`

The architecture has no `@support_torch_compile`, and on ROCm `default_breakable_cudagraph_architectures()` returns an empty set whatever the architecture, so breakable CUDA graphs are never auto-enabled here. The default `FULL_AND_PIECEWISE` therefore has no piecewise path and dies at capture:

```
RuntimeError: DeepseekV41ForCausalLM: piecewise CUDA graphs
(cudagraph_mode=FULL_AND_PIECEWISE) unavailable, model is not torch-compiled
and breakable CUDA graph is off.
```

Asking for plain `FULL` does not help either — the attention backend reports `AttentionCGSupport.UNIFORM_BATCH`, so it is downgraded straight back:

```
WARNING compilation.py CUDAGraphMode.FULL is not supported with
DeepseekV41ROCMAiterSparseSWABackend backend (support:
AttentionCGSupport.UNIFORM_BATCH); setting cudagraph_mode=FULL_AND_PIECEWISE
```

`DeepseekV41ForCausalLM` *is* listed in `DEFAULT_BREAKABLE_CUDAGRAPH_ARCHITECTURES`, so on CUDA `_maybe_enable_breakable_cudagraph` sets `VLLM_USE_BREAKABLE_CUDAGRAPH=1` for it automatically and the piecewise path exists. The ROCm empty-set override is what removes it, and it is the only reason this fails here.

The `VLLM_USE_BREAKABLE_CUDAGRAPH=1` override the error message suggests does work, but it is slower. Fresh server per arm, in=1024 out=256 at concurrency 1:

| cudagraph_mode | startup | median TTFT | median TPOT |
|---|---|---:|---:|
| `FULL_AND_PIECEWISE` (default) | fails at capture | — | — |
| `FULL_AND_PIECEWISE` + `VLLM_USE_BREAKABLE_CUDAGRAPH=1` | starts | 137.74 ms | 14.81 ms |
| **`FULL_DECODE_ONLY`** | starts | 135.52 ms | **14.64 ms** |

One run per arm, so take the 1.2% TPOT gap as directional rather than settled. It agrees in sign with the ROCm branch comment in `vllm/config/vllm.py`, which says breakable graphs currently regress performance there.

This is ROCm-specific: on CUDA the architecture auto-enables breakable graphs, so the raise does not fire there and the NVIDIA entries in `meta.hardware` need no change.

### Why `--no-async-scheduling`

With graphs enabled and async scheduling at its default, the engine segfaults on the **3rd decode token of the first request**, deterministically. Filed upstream as vllm-project/vllm#56347 with a minimal reproducer; this flag should be dropped from the recipe once that is fixed. Since filing it I have not been able to reproduce it: nine runs of that reproducer across three builds all passed, including on `vllm/vllm-openai-rocm:deepseekv41-flash-0909` with async scheduling left on. I cannot pin the overlay revision I originally tested, so I have kept the flag rather than drop it on a negative result, and recorded the attempts on the issue.

Running eager (`cudagraph_mode=NONE`) avoids both problems but costs **~3.5×** on decode — 149 ms/token versus 43 ms at concurrency 1 — so it is not a reasonable default.

## Test Plan

The recipe as composed with this change (TP4, one MI355X tray):

```bash
docker run --rm --device=/dev/kfd --device=/dev/dri --group-add video \
  --security-opt seccomp=unconfined --ipc=host --network=host \
  -v ~/.cache/huggingface:/root/.cache/huggingface \
  -e VLLM_ENGINE_READY_TIMEOUT_S=3600 \
  -e VLLM_USE_RUST_FRONTEND=1 \
  -e VLLM_ROCM_USE_AITER=1 \
  -e VLLM_ROCM_USE_AITER_MOE=1 \
  vllm/vllm-openai-rocm:deepseekv41-flash-0909 \
  deepseek-ai/DeepSeek-V4.1-Flash \
  --tokenizer-mode deepseek_v41 \
  --tensor-parallel-size 4 \
  --gpu-memory-utilization 0.9 \
  --moe-backend aiter_triton_mxfp4_bf16 \
  --compilation-config '{"cudagraph_mode":"FULL_DECODE_ONLY"}' \
  --no-async-scheduling \
  --tool-call-parser deepseek_v41 --enable-auto-tool-choice \
  --reasoning-parser deepseek_v41 \
  --mm-encoder-tp-mode data
```

The recipe's own verification:

```bash
curl http://localhost:8000/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{"model":"deepseek-ai/DeepSeek-V4.1-Flash",
       "messages":[{"role":"user","content":"What is 17*19? Return only the integer."}]}'
```

Throughput:

```bash
vllm bench serve --backend openai --base-url http://localhost:8000 \
  --model deepseek-ai/DeepSeek-V4.1-Flash \
  --tokenizer /path/to/DeepSeek-V4.1-Flash --tokenizer-mode deepseek_v41 \
  --dataset-name random --random-input-len 1024 --random-output-len 1024 \
  --num-prompts 32 --max-concurrency 8 --ignore-eos
```

What I actually ran differs in one respect, because the AMD image is not published (see the note at the end): `vllm/vllm-openai-rocm:nightly-2a02f6ef` with a python overlay from vllm-project/vllm#56214, `VLLM_USE_RUST_FRONTEND=0` (the image's prebuilt Rust frontend predates the V4.1 renderer), and the local model path in place of the HF id. Every flag in this PR was exercised exactly as written above.

## Result

Benchmarked with the command above at four concurrency levels, 1k in / 1k out:

| concurrency | output tok/s | median TTFT | median TPOT | failed requests |
|---|---:|---:|---:|---:|
| 1 | 23.21 | 206 ms | 42.90 ms | 0 |
| 8 | 172.88 | 742 ms | 44.55 ms | 0 |
| 32 | 548.66 | 1826 ms | 56.27 ms | 0 |
| 64 | 988.24 | 1828 ms | 62.19 ms | 0 |

Correctness: the recipe's own check returns `323` for `17*19`.

## Note on the AMD image

Separate from this change, and not addressed here: `docker_image.amd` names `vllm/vllm-openai-rocm:deepseekv41-flash-0909`, which is not published — that tag exists only on the CUDA repo `vllm/vllm-openai`, and `vllm-openai-rocm` currently has no `deepseek*` tag. My testing therefore used `vllm/vllm-openai-rocm:nightly` plus a source overlay from vllm-project/vllm#56214, which is what supplies DeepSeek-V4.1 on ROCm. Flagging it because `hardware.mi350x: verified` is not reproducible from the recipe as written until that image exists.

### DIFF
--- a/models/deepseek-ai/DeepSeek-V4.1-Flash.yaml
+++ b/models/deepseek-ai/DeepSeek-V4.1-Flash.yaml
@@ -97,6 +97,10 @@ hardware_overrides:
       - "0.9"
       - "--moe-backend"
       - "aiter_triton_mxfp4_bf16"
+      # Only working graph mode on ROCm; async scheduling crashes it (vllm#56347).
+      - "--compilation-config"
+      - '{"cudagraph_mode":"FULL_DECODE_ONLY"}'
+      - "--no-async-scheduling"
     extra_env:
       VLLM_ROCM_USE_AITER: "1"
       VLLM_ROCM_USE_AITER_MOE: "1"


JohnQinAMD:
Correcting two claims in the description, and adding the measurement I said I had not taken.

**What was wrong.** I wrote that `DeepseekV41ForCausalLM` is not in `DEFAULT_BREAKABLE_CUDAGRAPH_ARCHITECTURES` and that breakable graphs are therefore not auto-enabled for it "on CUDA either". Both are wrong: the architecture has been in that set since #56214 added the model, and `_maybe_enable_breakable_cudagraph` sets `VLLM_USE_BREAKABLE_CUDAGRAPH=1` for it automatically on CUDA. The only reason the default mode fails here is the ROCm branch of `default_breakable_cudagraph_architectures()`, which returns an empty set whatever the architecture.

**What I have now measured.** I had left the `VLLM_USE_BREAKABLE_CUDAGRAPH=1` override untested. It does work, and it is slower — fresh server per arm, in=1024 out=256 at concurrency 1:

| cudagraph_mode | startup | median TTFT | median TPOT |
|---|---|---:|---:|
| `FULL_AND_PIECEWISE` (default) | fails at capture | — | — |
| `FULL_AND_PIECEWISE` + `VLLM_USE_BREAKABLE_CUDAGRAPH=1` | starts | 137.74 ms | 14.81 ms |
| **`FULL_DECODE_ONLY`** | starts | 135.52 ms | **14.64 ms** |

I checked the resolved mode in each server's log rather than trusting a clean startup, since a silent downgrade would look the same from outside.

The change this PR makes is unaffected: `FULL_DECODE_ONLY` is still the mode to pin, and it is now the faster of the two that run rather than merely the only one. One run per arm, so treat the 1.2% as directional.

JohnQinAMD:
Closing this: on current `main` the recipe runs as published on
`vllm/vllm-openai-rocm:nightly-eed1f3d0c6043bd494424a22443ee198dd56f657`, so neither of the
two changes here is needed any more.

PR #958, opened 2026-09-11
Enable GLM-5.3-Flash MTP on MI355X

jamesETsmith:
# Summary

Replace the obsolete unsupported warning about MTP in the GLM-5.3-Flash MI355X recipe example now that vLLM's ROCm sparse-MLA dispatch fix has merged (see https://github.com/vllm-project/vllm/pull/55239)

# Testing
Validation: `node scripts/build-recipes-api.mjs`


esmeetu:
lgtm!

PR #960, opened 2026-09-12
Update K2-Horizon context length arguments

tanyuqian:
Removes `--max-model-len 131072` from base arguments and launch examples for all K2-Horizon models except 0.9B to prevent possible confusion about their supported context lengths.


esmeetu:
lgtm.

PR #976, opened 2026-09-15
[ROCm] Add AMD DeepSeek-V4-Flash-Vision-Exp recipe

indivats:
Adds a verified AMD ROCm path to the DeepSeek-V4-Flash-Vision-Exp recipe, built on [vllm-project/vllm#55107](https://github.com/vllm-project/vllm/pull/55107), which merged ROCm vision support for this checkpoint.
**Config:**
- `TP=2` is the pinned default for the long-context-agentic layout. A 32K-in/256-out /concurrency-4 sweep put TP2 ahead of TP4 (43.2 vs 42.6 output tok/s, 91 vs 92.5 ms TPOT). TP4 is verified and available.
- `AITER` is mandatory: the sparse-attention indexer's ROCm path only runs on AITER.
- Breakable CUDA graphs are required: the multimodal wrapper isn't torch-compiled, and default `FULL_AND_PIECEWISE` capture can't attach without them.
- The V2 model runner is required for DSpark, harmless for plain serving.
- The Strategy selector defaults to TP+EP (`--enable-expert-parallel`); both TP+EP and TP were benchmarked back-to-back on MI350X and came out equivalent (~43.7 tok/s / ~89 ms TPOT either way, at 32K in / 256 out / conc 4) with identical text and image outputs.


indivats:
@hongxiayang could you please take a look? thanks

hongxiayang (approved):
lgtm. Thanks @indivats

PR #979, opened 2026-09-15
GLM 5.3 NVFP4 recipe for DGX Station

prodney-dev:
Add support for CPU offload handling. DGX Station has unified system memory which can be used to extend the VRAM (252GB) up to 748GB of unified system memory. But we cannot arbitrarily increase the VRAM in the hardware definition because the CPU offload recipe is different from the case where all weights+headroom fit in VRAM.


models/zai-org/GLM-5.3.yaml
```diff
@@ -92,7 +99,36 @@ variants:
...
+          - "262144"
+          - "--max-num-seqs"
+          - "8"
+          - "--offload-backend"
+          - "uva"
+          - "--cpu-offload-gb"
+          - "222"
+          - "--cpu-offload-params"
+          - "w13_weight"
+          - "w2_weight"
```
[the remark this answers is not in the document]
prodney-dev: Overrides are not expected for this argument.

models/zai-org/GLM-5.3.yaml
```diff
@@ -289,6 +328,38 @@ guide: |
     --served-model-name glm-5.3-nvfp4
   ```
 
+  ### NVFP4 on DGX Station (GB300, CPU weight offload)
```
esmeetu: those command lines seems redundant.

jmeadlock:
+1 on this schema change from a second Station data point, in case it helps the review.

We have DeepSeek-V4.1-Flash serving at full 1M context on one DGX Station GB300 with the same mechanism this PR is built around — `--offload-backend uva --cpu-offload-gb 60 --cpu-offload-params routed_experts.w13_weight routed_experts.w2_weight` plus `--engram-config '{"cpu_offload":true}'` (the 189 GiB Engram table in pinned host memory). Flag-only, on the current nightly (`dee37d89`, vLLM `v0.29.1rc1.dev347`), measured 2026-09-20 on two boots: **90.5 tok/s single-stream prose · 302 tok/s at 8 streams · ~420 at 16 · KV pool 3.09M tokens**, 7.5-min bind on an autotune-cache hit. Receipts, launch script and the same-window nvfp4-vs-fp8 KV comparison: https://github.com/J-M-Recipes/recipes/tree/main/recipes/dgx-station-gb300/deepseek-v4.1-flash-vllm-uva-dspark/results/2026-09-19-overnight-ksched-nightly-adaptive/T3c

Today `DeepSeek-V4.1-Flash.yaml` has `vram_minimum_gb: 614` and no Station row, so the builder hides it on `dgx_station_gb300` (`scalable: false`) — exactly the case `hardware_overrides.<gpu>.vram_minimum_gb` fixes. I'd like to file that row on top of this PR once it lands rather than duplicate the schema work.

One question for whoever owns the schema: V4.1-Flash has only a `default` variant, and the exact-GPU `hardware_overrides` keyspace is documented as non-default variants only. For a single-variant recipe, is the intended pattern a named variant (e.g. `dgx-station-uva-offload`, same `model_id`, its own `vram_minimum_gb`), or would you rather relax that rule? Happy to go either way.

models/zai-org/GLM-5.3.yaml
prodney-dev: Removed

prodney-dev:
@esmeetu , I have resolved your comment, can you take a look? Thanks

esmeetu:
Thanks!

jmeadlock:
Filed the V4.1-Flash Station row on top of this: #1015. Went with `variants.default.hardware_overrides.dgx_station_gb300` since the merged code has no default-variant gate on the exact-GPU keyspace and the file already uses it for h100 — flagged in the PR in case you'd rather keep the doc rule strict.

PR #984, opened 2026-09-16
fix(kimi-k3): enable dspark spec decoding with pipeline parallel

namgyu-youn:
## Summary

Follow-up https://github.com/vllm-project/recipes/pull/895#issuecomment-5652251914

DSpark speculative decoding was gated off, because of upstream issue. That gap has since been fixed upstream (https://github.com/vllm-project/vllm/pull/50514) and validated end-to-end on Kimi-K3 + DSpark at TP8×PP2, so the strategy restriction is no longer needed and the recipe's minimum vLLM version is bumped to match.


esmeetu:
Thanks!

